PR #2, opened 2023-05-29
feat: add logic of uniswap v2 + elastic rate calculator

piavgh:


pkg/source/pool/iface.go
```diff
@@ -0,0 +1,43 @@
...
+	) (*CalcAmountOutResult, error)
+	UpdateBalance(params UpdateBalanceParams)
+	CanSwapTo(address string) []string
+	GetTokens() []string
+	GetAddress() string
+	GetExchange() string
+	GetType() string
+	GetTokenIndex(address string) int
+	GetLpToken() string
+	GetMidPrice(tokenIn string, tokenOut string, base *big.Int) *big.Int
```
hiepnv90, line 39: If this method was not used by other service, we should consider to delete it, we also need to consider to delete other methods in order to make the interface simpler.

pkg/source/pool/iface.go
```diff
@@ -0,0 +1,43 @@
...
+type IPoolSimulator interface {
+	// CalcAmountOut amountOut, fee, gas
+	// DO NOT FUCKING MODIFY THE POOL whilst calculating amount out. Call UpdateBalance when you need to.
+	CalcAmountOut(
+		tokenAmountIn TokenAmount,
+		tokenOut string,
+	) (*CalcAmountOutResult, error)
+	UpdateBalance(params UpdateBalanceParams)
+	CanSwapTo(address string) []string
+	GetTokens() []string
```
hiepnv90, line 33: It looks like this 2 methods are duplicated, we can find swapTo tokens from all tokens and input token address. I think we should write a helper function for `CanSwapTo` instead of make it a method of the interface.

pkg/source/pool/iface.go
hiepnv90: Example:

```go
func CanSwapTo(p IPoolSimulator, tokenIn string) []string {
    tokens := p.GetTokens()
    tokenOuts := make([]string, 0, len(tokens))
    for _, token := range tokens {
        if token != tokenIn {
            tokenOuts = append(tokenOuts, token)
        }
    }

    return tokenOuts
}
```

pkg/source/pool/iface.go
piavgh: Explanation here: https://team-kyber.slack.com/archives/C04SNUEPC4S/p1685602202537059?thread_ts=1685359191.861149&cid=C04SNUEPC4S

hiepnv90: Can we move this method to `pool.Pool` instead of re-implementing for every pools? Other liquidity source with different logic can overwrite it?

PR #9, opened 2023-06-02
[AG-347] copy makerpsm from router-service and add test

it4rb:
https://team-kyber.atlassian.net/browse/AG-347
(testcases are from https://github.com/makerdao/dss-psm/blob/master/src/tests/psm.t.sol)

### DIFF
--- a/pkg/source/makerpsm/pool_simulator.go
+++ b/pkg/source/makerpsm/pool_simulator.go
@@ -105,8 +105,6 @@ func (p *PoolSimulator) UpdateBalance(params pool.UpdateBalanceParams) {
 	p.PSM.updateBalanceSellingGem(output.Amount)
 }
 
-func (p *PoolSimulator) GetLpToken() string { return "" }
-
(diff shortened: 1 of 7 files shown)


pkg/source/makerpsm/pool_simulator.go
```diff
@@ -0,0 +1,112 @@
...
+func (p *PoolSimulator) UpdateBalance(params pool.UpdateBalanceParams) {
+	input, output := params.TokenAmountIn, params.TokenAmountOut
+	if strings.EqualFold(input.Token, DAIAddress) {
+		p.PSM.updateBalanceBuyingGem(input.Amount)
+	}
+
+	p.PSM.updateBalanceSellingGem(output.Amount)
+}
+
+func (p *PoolSimulator) GetLpToken() string { return "" }
```
piavgh, line 108: this method is not needed anymore

piavgh (approved):

pkg/source/makerpsm/pool_simulator.go
it4rb, line 108:
```suggestion
```

PR #12, opened 2023-06-05
[AG-330] [AG-331] add balancer dex (stable and weighted)

it4rb:
https://team-kyber.atlassian.net/browse/AG-330
https://team-kyber.atlassian.net/browse/AG-331

### DIFF
--- a/pkg/source/balancer/stable/math.go
+++ b/pkg/source/balancer/stable/math.go
@@ -1,0 +1,173 @@
+package balancerstable
+
+import (
+	"math/big"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+)
+
+var One *big.Int
+var AmpPrecision = bignumber.NewBig10("1000")
+
+func init() {
+	One = new(big.Int).Set(bignumber.BONE)
+}
+
+func _upscale(amount *big.Int, scalingFactor *big.Int) *big.Int {
+	return mulDown(amount, scalingFactor)
+}
+
+func _computeScalingFactor(tokenDecimals uint) *big.Int {
+	var decimalsDiff = 36 - tokenDecimals
+	return bignumber.TenPowInt(uint8(decimalsDiff))
+}
+
+func mulDown(a *big.Int, b *big.Int) *big.Int {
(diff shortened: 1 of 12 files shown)


pkg/source/balancer/stable/math.go
```diff
@@ -0,0 +1,173 @@
...
+	"math/big"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+)
+
+var One *big.Int
+var AmpPrecision = bignumber.NewBig10("1000")
+
+func init() {
+	One = bignumber.BONE
```
piavgh, line 13: to avoid mutating the BONE via One, I think we should do `One = new(big.Int).SetInt(bignumber.BONE)`

pkg/source/balancer/stable/pool_simulator.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (t *StablePool) GetMetaInfo(tokenIn string, tokenOut string) interface{} {
+	return balancer.Meta{
+		VaultAddress: t.VaultAddress,
+		PoolId:       t.PoolId,
+	}
+}
+
+func (t *StablePool) GetLpToken() string {
```
piavgh, line 126: this method is not used anywhere so we can remove it

pkg/source/balancer/weighted/pool_simulator.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (t *WeightedPool2Tokens) GetMetaInfo(tokenIn string, tokenOut string) interface{} {
+	return balancer.Meta{
+		VaultAddress: t.VaultAddress,
+		PoolId:       t.PoolId,
+	}
+}
+
+func (t *WeightedPool2Tokens) GetLpToken() string {
```
piavgh, line 130: same here, we can remove this method

piavgh:
LGTM, but please help to run this code with router-service to make sure that we don't have cycle import issue

piavgh (approved):

it4rb:
@piavgh : I've checked with router-service, we can add these 2 modules in router-service and compile successfully, so I'll merge this PR

PR #14, opened 2023-06-05
create workflow

guiltylotus:


.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+          git tag --annotate --message "dmm-aggregator-backend $VERSION_TAG" "$VERSION_TAG"
+          git push origin "refs/tags/$VERSION_TAG"
+
+      - name: Create release
+        uses: softprops/action-gh-release@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          tag_name: ${{ env.VERSION_TAG }}
+          prerelease: false
+          name: "Pool Service ${{ env.VERSION_TAG }}"
```
lehainam-dev, line 123: This may need to be change.

guiltylotus: Updated ^^

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version'
+        type: string
+        required: true
+
+jobs:
+  prepare:
+    runs-on: [ self-hosted ]
```
it4rb, line 13:
```suggestion
    runs-on: [ ubuntu-22.04 ]
```
we cannot use self-hosted runner for public repo

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Move Docker cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+
+  release:
+    needs:
+      - prepare
+      - docker
+    runs-on: [ self-hosted ]
```
it4rb, line 92:
```suggestion
    runs-on: [ ubuntu-22.04 ]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Setup Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Create tag
+        run: |
+          git tag -d "$VERSION_TAG" 2> /dev/null || echo "Release tag '$VERSION_TAG' does NOT exist"
+          git tag --annotate --message "dmm-aggregator-backend $VERSION_TAG" "$VERSION_TAG"
```
it4rb, line 114: um maybe we need to change this message?

.github/workflows/release.yml
```diff
@@ -0,0 +1,123 @@
...
+        run: echo "::set-output name=value::$(date +%FT%T%z)"
+
+  docker:
+    needs:
+      - prepare
+    env:
+      SERVICE: kyberswap-dex-lib
+      VERSION_TAG: ${{ needs.prepare.outputs.version_tag }}
+      BUILD_DATE: ${{ needs.prepare.outputs.build_date }}
+    runs-on: [ self-hosted ]
```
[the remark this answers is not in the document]
guiltylotus: Why can not use `self-hosted` here?

.github/workflows/release.yml
guiltylotus: Updated ^^

.github/workflows/release.yml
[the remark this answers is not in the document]
it4rb: um this is a public repository, while our self-hosted runner is private, so github won't be able to run this action
other public repo need to use ubuntu as well, for example https://github.com/KyberNetwork/evmlistener/commit/285c591a611dbd4b38526528962a0f9853a71a76

it4rb (approved):

.github/workflows/release.yml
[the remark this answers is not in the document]
guiltylotus: Updated

.github/workflows/release.yml
guiltylotus: Updated

.github/workflows/release.yml
guiltylotus: Updated

PR #16, opened 2023-06-06
[AG-334] add curve-base simulation logic

it4rb:


pkg/source/curve/base/pool_simulator.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		if t.Info.Tokens[i] == output.Token {
+			t.Info.Reserves[i] = new(big.Int).Sub(t.Info.Reserves[i], outputAmount)
+		}
+	}
+}
+
+func (t *PoolBaseSimulator) GetLpToken() string {
+	return t.LpToken
+}
```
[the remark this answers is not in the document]
it4rb: that CanSwapTo is not in agreement with CalcAmountOut (https://github.com/KyberNetwork/router-service/blob/b2d2b79094e56cb6614b17c2dae1f66b72a6a044/internal/pkg/core/curve-base/pool.go#L99), in CalcAmountOut we can only swap the pool tokens, not its lpToken

so I think we should limit it to what the pool can actually swap (by removing CanSwapTo and using default logic)

(Hank and I already discussed this somewhere on slack but unfortunately I don't remember where)

pkg/source/curve/base/math.go
```diff
@@ -0,0 +1,464 @@
...
+	var numTokensBI = big.NewInt(int64(numTokens))
+	var prevD *big.Int
+	var d = new(big.Int).Set(s)
+	var nA = new(big.Int).Mul(a, numTokensBI)
+	for i := 0; i < MaxLoopLimit; i++ {
+		var dP = new(big.Int).Set(d)
+		for j := 0; j < numTokens; j++ {
+			dP = new(big.Int).Div(
+				new(big.Int).Mul(dP, d),
+				new(big.Int).Mul(xp[j], numTokensBI),
```
piavgh, line 103: newest code: https://github.com/KyberNetwork/router-service/blob/40b9d45e99b213e31bc4eaaf75f325c185129b2d/internal/pkg/core/curve-base/math.go#L104

pkg/source/curve/type.go
```diff
@@ -174,3 +174,13 @@ type PoolTricryptoExtra struct {
 	AdjustmentStep      string   `json:"adjustmentStep"`
 	MaHalfTime          string   `json:"maHalfTime"`
 }
+
+type Gas struct {
```
piavgh, line 178: I think we have to move this struct into `base` package

Because other Curve pool types have different Gas struct

pkg/source/curve/type.go
```diff
@@ -174,3 +174,13 @@ type PoolTricryptoExtra struct {
...
+	Exchange int64
+}
+
+type Meta struct {
```
piavgh, line 182: if we leave this Meta struct here, it's better to check if we have cycle import issue

pkg/source/curve/base/pool_simulator.go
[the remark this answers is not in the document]
piavgh: @hanguyenk : if you have the link to that discussion, please send it here. We changed the old implementation, so I think we should have a strong reason to do so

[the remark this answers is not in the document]
it4rb: @piavgh : here it is https://team-kyber.slack.com/archives/C054H6WG7EV/p1686886822621859?thread_ts=1686885353.928249&cid=C054H6WG7EV

pkg/source/curve/type.go
it4rb: I don't think we'll have problem here, because router-service will import the simulator in `base` package directly, not through the `curve` package

pkg/source/curve/base/pool_simulator.go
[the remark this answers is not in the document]
piavgh: ok, so I think it's good to remove lpToken

pkg/source/curve/base/pool_simulator.go
```diff
@@ -0,0 +1,168 @@
...
```
it4rb, line 158:
```suggestion
```

pkg/source/curve/base/pool_simulator_test.go
```diff
@@ -0,0 +1,86 @@
...
+			out, err := p.CalcAmountOut(pool.TokenAmount{Token: tc.in, Amount: big.NewInt(tc.inAmount)}, tc.out)
+			require.Nil(t, err)
+			assert.Equal(t, big.NewInt(tc.expectedOutAmount), out.TokenAmountOut.Amount)
+			assert.Equal(t, tc.out, out.TokenAmountOut.Token)
+			assert.Equal(t, big.NewInt(tc.expectedFeeAmount), out.Fee.Amount)
+		})
+	}
+}
+
+func TestCalcAmountOut_futureA(t *testing.T) {
```
piavgh, line 61: This function name is a bit confusing (I don't understand what it does with the futureA).
It's better if you can have a more detailed name for it

piavgh (approved):

pkg/source/curve/base/pool_simulator_test.go
it4rb, line 61:
```suggestion
func TestCalcAmountOut_interpolate_from_initialA_and_futureA(t *testing.T) {
```

PR #18, opened 2023-06-06
[AG-333] add curve-aave simulation logic

it4rb:
(this PR depends on https://github.com/KyberNetwork/kyberswap-dex-lib/pull/16)

### DIFF
--- a/pkg/source/curve/aave/pool_simulator.go
+++ b/pkg/source/curve/aave/pool_simulator.go
@@ -1,0 +1,315 @@
+package aave
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+	"strings"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/curve"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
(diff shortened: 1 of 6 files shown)


pkg/source/curve/aave/pool_simulator.go
```diff
@@ -0,0 +1,315 @@
...
+	"encoding/json"
+	"fmt"
+	"math/big"
+	"strings"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/curve"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+	constant "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
```
piavgh, line 13: one of these 2 lines is redundant

piavgh: it's better if we can keep our code consistent so that it will not make new developers confused

piavgh (approved):

pkg/source/curve/aave/pool_simulator.go
it4rb: yes, if we've aliased once before then vscode will automatically alias when pasting new code (copied from router-service), sorry I've should have checked this

PR #20, opened 2023-06-09
[AG-364] integrate Lido eth-steth

it4rb:
For Lido ETH->stETH we need to get 2 reserved amounts (totalShares and totalPooledEther)

Formula in their contract code:
```
/**
     * @return the amount of shares that corresponds to `_ethAmount` protocol-controlled Ether.
     */
    function getSharesByPooledEth(uint256 _ethAmount) public view returns (uint256) {
        return _ethAmount
            .mul(_getTotalShares())
            .div(_getTotalPooledEther());
    }
```

### DIFF
--- a/pkg/source/lido-steth/pool_tracker.go
+++ b/pkg/source/lido-steth/pool_tracker.go
@@ -37,7 +37,6 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
 	}
 
 	p.Reserves = reserves
-	p.Extra = "{}"
 	p.Timestamp = time.Now().Unix()
 
 	log.Infof("[Lido-stETH] Finish getting new state of pool")
--- a/pkg/source/lido-steth/pool_tracker.go
+++ b/pkg/source/lido-steth/pool_tracker.go
@@ -1,0 +1,78 @@
+package lido_steth
+
+import (
+	"context"
+	"math/big"
+	"time"
+
+	"github.com/KyberNetwork/ethrpc"
+	"github.com/KyberNetwork/logger"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+)
+
+type PoolTracker struct {
+	ethrpcClient *ethrpc.Client
+}
+
+func NewPoolTracker(ethrpcClient *ethrpc.Client) *PoolTracker {
+	return &PoolTracker{
+		ethrpcClient: ethrpcClient,
+	}
+}
+
+func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entity.Pool, error) {
+	log := logger.WithFields(logger.Fields{
+		"poolAddress": p.Address,
+	})
+
+	log.Infof("[Lido-stETH] Start getting new pool's state")
+
+	reserves, err := d.getPoolReserves(ctx, p)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/source/lido/pools/ethereum.json
```diff
@@ -2,6 +2,7 @@
   {
     "id": "0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0",
     "name": "wstETH",
+    "type": "lido",
```
piavgh, line 5: to make it similar to the one below, can we set this as `lido-steth-wsteth`? Is there any problem by doing so?

it4rb: um I tried that before but the existed Lido pool didn't get updated so I decided to keep it the same

pkg/source/lido-steth/pool_tracker.go
```diff
@@ -0,0 +1,78 @@
...
+	reserves, err := d.getPoolReserves(ctx, p)
+	if err != nil {
+		log.WithFields(logger.Fields{
+			"error": err,
+		}).Error("failed to getPoolReserves")
+		return entity.Pool{}, err
+	}
+
+	p.Reserves = reserves
+	p.Extra = "{}"
```
hanguyenk, line 40: [small] I think we don't need to specify extra in this case

pkg/source/lido-steth/pools_list_updater.go
```diff
@@ -0,0 +1,119 @@
...
+
+func NewPoolsListUpdater(cfg *Config) *PoolsListUpdater {
+	return &PoolsListUpdater{
+		config: cfg,
+	}
+}
+
+func (d *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte) ([]entity.Pool, []byte, error) {
+	if d.hasInitialized {
+		logger.Info("skip since pool has been initialized")
```
hanguyenk, line 28: [small] I think this log should be in debug level

pkg/source/lido-steth/pools_list_updater.go
```diff
@@ -0,0 +1,119 @@
...
+	"time"
+
+	"github.com/KyberNetwork/logger"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+)
+
+type PoolsListUpdater struct {
+	config         *Config
+	hasInitialized bool
```
hanguyenk, line 17: If the runner pod is restarted, we will initialize pools all over again? 
Personally I think this behavior is fine as long as it doesn't overwrite the existing ones in Redis.

it4rb: yes, we'll reinitialize on restart, and pool-service will skip existed pool without writing to redis https://github.com/KyberNetwork/pool-service/blob/main/internal/runner/pools_list_updater.go#L72

pkg/source/lido-steth/pool_tracker.go
it4rb, line 40:
```suggestion
```

pkg/source/lido-steth/pools_list_updater.go
it4rb, line 28:
```suggestion
		logger.Debug("skip since pool has been initialized")
```

hanguyenk (approved):

PR #25, opened 2023-06-14
[AG-355] add Velodrome simulation logic

it4rb:


pkg/source/velodrome/pool_simulator.go
```diff
@@ -0,0 +1,125 @@
...
+		ReserveUsd: entityPool.ReserveUsd,
+		SwapFee:    swapFee,
+		Exchange:   entityPool.Exchange,
+		Type:       entityPool.Type,
+		Tokens:     tokens,
+		Reserves:   reserves,
+		Checked:    false,
+	}
+
+	staticExtra, err := extractStaticExtra(entityPool.StaticExtra)
```
piavgh, line 47: because the function extractStaticExtra is used in both pool_simulator and pool_tracker, I think we should move it to the helper.go (or something similar).
It makes the package look more organized

piavgh (approved):

piavgh:
this PR has conflicts, so please rebase it onto the main also

PR #27, opened 2023-06-15
[AG-335] add curve compound simulation logic

it4rb:


pkg/source/curve/compound/math.go
```diff
@@ -0,0 +1,175 @@
+package compound
+
+import (
+	"math/big"
+
+	constant "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+	utils "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+)
+
+var FeeDenominator = utils.NewBig10("10000000000")
```
piavgh, line 10: I think we can move this to `constant.go`

piavgh (approved):

PR #30, opened 2023-06-19
[AG-339] add curve-two simulation logic

it4rb:


pkg/source/curve/two/constant.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/curve"
+	constant "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+	utils "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+)
+
+var (
+	DefaultGas       = curve.Gas{Exchange: 220000}
+	MinGamma         = utils.NewBig10("10000000000")
+	MaxGamma         = new(big.Int).Mul(big.NewInt(5), utils.NewBig10("10000000000000000"))
```
lehainam-dev, line 14: MaxGamma here should be 2 * 10000000000000000.

lehainam-dev, line 14: I think [MaxGamma should be 2 * 10**16](https://github.com/curvefi/curve-crypto-contract/blob/master/contracts/two/CurveCryptoSwap2.vy#L156C1-L156C10). I realized that when adding contract's safe-check in `newtonY`.

```suggestion
	MaxGamma         = new(big.Int).Mul(big.NewInt(2), utils.NewBig10("10000000000000000"))
```

pkg/source/curve/two/math.go
```diff
@@ -0,0 +1,762 @@
...
+					return nil, errors.New("unsafe values x[i]")
+				}
+			}
+			return D, nil
+		}
+	}
+	return nil, errors.New("did not converge")
+}
+
+func newtonY(ann *big.Int, gamma *big.Int, x []*big.Int, D *big.Int, i int) (*big.Int, error) {
```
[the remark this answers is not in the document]
it4rb: thanks! I've added the check and also copied another missing fix for panic as well

piavgh (approved):

PR #34, opened 2023-06-19
[AG-336] add curve-meta simulation logic

it4rb:


pkg/source/curve/meta/math.go
```diff
@@ -0,0 +1,367 @@
...
+	var numTokensBI = big.NewInt(int64(numTokens))
+	var prevD *big.Int
+	var d = new(big.Int).Set(s)
+	var nA = new(big.Int).Mul(a, numTokensBI)
+	for i := 0; i < MaxLoopLimit; i++ {
+		var dP = new(big.Int).Set(d)
+		for j := 0; j < numTokens; j++ {
+			dP = new(big.Int).Div(
+				new(big.Int).Mul(dP, d),
+				new(big.Int).Mul(xp[j], numTokensBI),
```
piavgh, line 41: Newest code: https://github.com/KyberNetwork/router-service/blob/fa4f6ac7688ad60171f50040abacdcac1cb6be04/internal/pkg/core/curve-meta/math.go#L42

pkg/source/curve/meta/pool_simulator.go
```diff
@@ -0,0 +1,279 @@
...
+	if outputIndex < 0 && baseOutputIndex >= 0 {
+		outputIndex = maxCoin + baseOutputIndex
+	}
+	if inputIndex >= 0 && outputIndex >= 0 {
+		// exchange_underlying
+		_, _ = t.ExchangeUnderlying(inputIndex, outputIndex, inputAmount)
+	}
+}
+
+func (t *Pool) GetLpToken() string {
```
piavgh, line 209: I guess we can remove this method

pkg/source/curve/meta/math.go
it4rb: hmm changing this will make the result in testcase differ than real result obtained from etherscan, so I think I'll add a check and return error if denominator is zero

https://github.com/curvefi/curve-contract/blob/d4e8589ac92c4019b3064b2f3a8a87dbc3281b46/contracts/pools/usdn/StableSwapUSDN.vy#L296

piavgh (approved):

PR #35, opened 2023-06-20
[AG-337] add curve-plain-oracle simulation logic

it4rb:


pkg/source/curve/plain-oracle/constant.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/curve"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+)
+
+const MaxLoopLimit = 256
+
+var (
+	DefaultGas     = curve.Gas{Exchange: 128000}
```
piavgh, line 11: I think it should have a separate Gas struct instead of using the same Gas struct from the `curve` package

pkg/source/curve/plain-oracle/math.go
```diff
@@ -0,0 +1,422 @@
...
+	var numTokensBI = big.NewInt(int64(numTokens))
+	var prevD *big.Int
+	var d = new(big.Int).Set(s)
+	var nA = new(big.Int).Mul(a, numTokensBI)
+	for i := 0; i < MaxLoopLimit; i++ {
+		var dP = new(big.Int).Set(d)
+		for j := 0; j < numTokens; j++ {
+			dP = new(big.Int).Div(
+				new(big.Int).Mul(dP, d),
+				new(big.Int).Mul(xp[j], numTokensBI),
```
piavgh, line 98: newest code: https://github.com/KyberNetwork/router-service/blob/40b9d45e99b213e31bc4eaaf75f325c185129b2d/internal/pkg/core/curve-plain-oracle/math.go#L100

piavgh (approved):

PR #36, opened 2023-06-20
[AG-344] add lido-wsteth simulation logic

it4rb:


pkg/source/lido/pool_simulator.go
```diff
@@ -0,0 +1,101 @@
+package lido
```
piavgh, line 1: I think we can change this package to `lido_wsteth` to make it similar to the package `lido_steth`
What do you think @it4rb @hanguyenk ?

piavgh (approved):

pkg/source/lido/pool_simulator.go
it4rb: I'm not very sure about that, if we decide to rename package then we might want to rename constants like dex's name/type as well, that would required cleaning up data in redis, it's not a big deal but I would like to avoid that if possible
anyway let me create a ticket and plan to that later

PR #44, opened 2023-06-23
add logic to classify curve crypto

guiltylotus:


pkg/source/curve/pool_lists_classifier.go
```diff
@@ -169,6 +177,9 @@ func (d *PoolsListUpdater) classifyCurveV2PoolTypes(
 
 	var poolTypes = make([]string, len(poolAddresses))
 	for i := range poolAddresses {
+		if gammaList[i] == nil {
+			continue
```
lehainam-dev, line 181: Maybe we should log the skip `poolAddress` here.

PR #46, opened 2023-06-26
feat: integrate ellipsis

lehainam-dev:
Related bug: https://team-kyber.atlassian.net/browse/AG-452

### DIFF
--- a/pkg/source/curve/constant.go
+++ b/pkg/source/curve/constant.go
@@ -103,4 +103,7 @@ var (
 	zero        int64 = 0
 )
 
-var skipInitFactoryDexes = []string{"ellipsis"}
+// skipInitFactoryDexes include dexes that don't have factory.
+var skipInitFactoryDexes = map[string]struct{}{
+	"ellipsis": {},
+}
--- a/pkg/source/curve/constant.go
+++ b/pkg/source/curve/constant.go
@@ -102,3 +102,5 @@ var (
 	emptyString       = ""
 	zero        int64 = 0
 )
+
+var skipInitFactoryDexes = []string{"ellipsis"}
--- a/pkg/source/curve/embed.go
+++ b/pkg/source/curve/embed.go
@@ -62,6 +62,11 @@ var optimismPoolsBytes []byte
 //go:embed pools/polygon.json
 var polygonPoolsBytes []byte
 
+// Ellipsis pool bytes
+
+//go:embed pools/ellipsis/bsc.json
+var ellipsisBscPoolsBytes []byte
+
 var bytesByPath = map[string][]byte{
 	"pools/arbitrum.json":  arbitrumPoolsBytes,
 	"pools/avalanche.json": avalanchePoolsBytes,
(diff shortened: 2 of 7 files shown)


pkg/source/curve/constant.go
```diff
@@ -102,3 +102,5 @@ var (
 	emptyString       = ""
 	zero        int64 = 0
 )
+
+var skipInitFactoryDexes = []string{"ellipsis"}
```
hanguyenk, line 106: 1. Please add a comment to explain the meaning of this list
2. I think a map would be more efficient in this case

pkg/source/curve/embed.go
```diff
@@ -62,11 +62,18 @@ var optimismPoolsBytes []byte
...
+
 var bytesByPath = map[string][]byte{
 	"pools/arbitrum.json":  arbitrumPoolsBytes,
 	"pools/avalanche.json": avalanchePoolsBytes,
 	"pools/ethereum.json":  ethereumPoolsBytes,
 	"pools/fantom.json":    fantomPoolsBytes,
 	"pools/optimism.json":  optimismPoolsBytes,
 	"pools/polygon.json":   polygonPoolsBytes,
+
+	"pools/ellipsis/bsc.json": ellipsisBscPoolsBytes,
```
hanguyenk, line 78: How about `bsc_ellipsis.json`?

pkg/source/curve/pools_list_updater.go
```diff
@@ -65,37 +67,39 @@ func (d *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte
...
-		newPoolLimitLeft = newPoolLimitLeft - (nextOffset - registryOrFactoryList[i].Offset)
-
-		for j := 0; j < len(poolAddresses); j++ {
-			poolTypeMap[poolTypes[j]] = append(poolTypeMap[poolTypes[j]], PoolAndRegistries{
-				PoolAddress:              poolAddresses[j],
-				RegistryOrFactoryABI:     &registryOrFactoryList[i].ABI,
-				RegistryOrFactoryAddress: &registryOrFactoryList[i].Address,
-			})
-		}
+	if !skipInitFactory(d.config.DexID) {
```
hanguyenk, line 70: please use early return here

it4rb (approved):

pkg/source/curve/embed.go
lehainam-dev: I follow the same convention with `saddle`, where each dex is grouped into a single folder under the `pools` folder.
I think I still keep `pools/ellipsis/bsc.json` to be consistent with others.

pkg/source/curve/constant.go
lehainam-dev: Fixed in the latest commit.

pkg/source/curve/pools_list_updater.go
lehainam-dev: Below the condition, there's a shared logic so because of that, I can't use early return here.

hanguyenk (approved):

PR #47, opened 2023-06-28
fix: curve incorrect pool classify

piavgh:
- Change `*abi.ABI` and `*string` to `abi.ABI` and `string` to avoid weird issue wrong ABI file => `get_underlying_coins not found`
- Stop using `logger.WithFields` because it logs incorrect position of code
- In case of Optimism, the `meta pools factory` address is equal to the `crypto pools factory` address
  - We should only process Curve Two and Curve Tricrypto for `crypto pools registry` and `crypto pools factory`
  - Thus, now I added `poolsSourceIndex`. 
    - 1 = `main registry`, 2 = `meta pools factory`, 3 = `crypto pools registry`, 4 = `crypto pools factory`
    - Only 3 and 4 should be applied for Curve Two and Curve Tricrypto

### DIFF
--- a/pkg/source/curve/pool_lists_classifier.go
+++ b/pkg/source/curve/pool_lists_classifier.go
@@ -181,6 +181,7 @@ func (d *PoolsListUpdater) classifyCurveV2PoolTypes(
 		} else if d.isTricrypto(coins[i]) {
 			poolTypes[i] = poolTypeTricrypto
 		} else {
+			logger.Infof("unsupported curve v2 pool: %s", poolAddresses[i].Hex())
 			poolTypes[i] = poolTypeUnsupported
 		}
 	}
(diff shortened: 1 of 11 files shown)


pkg/source/curve/pool_lists_classifier.go
```diff
@@ -171,8 +178,10 @@ func (d *PoolsListUpdater) classifyCurveV2PoolTypes(
 	for i := range poolAddresses {
 		if d.isTwo(coins[i]) {
 			poolTypes[i] = poolTypeTwo
-		} else {
+		} else if d.isTricrypto(coins[i]) {
 			poolTypes[i] = poolTypeTricrypto
+		} else {
+			poolTypes[i] = poolTypeUnsupported
```
lehainam-dev, line 184: Should we log here, just to check if there is an abnormal pool landed on this case?

piavgh: good point, I added a commit to log it

lehainam-dev (approved):
I approve. Run locally on Optimism & Arbitrum and it works fine.

PR #48, opened 2023-06-28
get lptoken totalSupply for oneswap

it4rb:
oneswap is using saddle pool simulator, which expect totalSupply (to be able to swap to/from lpToken)

### DIFF
--- a/pkg/source/oneswap/pool_tracker.go
+++ b/pkg/source/oneswap/pool_tracker.go
@@ -101,11 +101,11 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
 		return entity.Pool{}, err
 	}
 
-	var reserves = make([]string, len(balances)+1)
-	for i := range balances {
-		reserves[i] = balances[i].String()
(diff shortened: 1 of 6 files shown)


pkg/source/oneswap/pool_tracker.go
```diff
@@ -72,10 +73,39 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
...
+			"error":       err,
+		}).Errorf("failed to aggregate pool total supply")
+		return entity.Pool{}, err
+	}
+
+	var reserves = make([]string, len(balances)+1)
 	for i := range balances {
 		reserves[i] = balances[i].String()
 	}
+	reserves[len(balances)] = totalSupply.String()
```
hanguyenk, line 108: might be a better implementation

```go
reserves := make([]string, 0, len(balances)+1)
for _, balance := range balances {
  reserves = append(reserves, balance.String())
}

reserves = append(reserves, totalSupply.String())
```

hanguyenk (approved):

PR #52, opened 2023-06-29
[AG-346] add madmex simulation

it4rb:


pkg/source/madmex/fast_price_feed_v1.go
```diff
@@ -35,3 +39,84 @@ const (
 	fastPriceFeedMethodV1Prices                    = "prices"
 	fastPriceFeedMethodV1VolBasisPoints            = "volBasisPoints"
 )
+
+func (pf *FastPriceFeedV1) GetPrice(token string, refPrice *big.Int, maximise bool) *big.Int {
+	fmt.Println("ffv1 getPrice", refPrice)
```
piavgh, line 44: I think we should not use fmt.Println, we should use logger.Debugf for something like this

pkg/source/madmex/fast_price_feed_v1.go
```diff
@@ -35,3 +39,84 @@ const (
...
+	if new(big.Int).SetInt64(time.Now().Unix()).Cmp(new(big.Int).Add(pf.LastUpdatedAt, pf.PriceDuration)) > 0 {
+		fmt.Println("ffv1 getPrice 0", refPrice)
```
piavgh, line 46: same for multiple other places

pkg/source/madmex/fast_price_feed_v1.go
it4rb: oops sorry, added those for debugging at router-service but forgot to remove before copying to dex-lib, will remove now

piavgh (approved):

PR #63, opened 2023-07-10
usdfi

tbotteam:


pkg/source/usdfi/config.go
```diff
@@ -0,0 +1,8 @@
+package usdfi
+
+type Config struct {
+	DexID          string  `json:"dexID"`
+	FactoryAddress string  `json:"factoryAddress"`
+	NewPoolLimit   int     `json:"newPoolLimit"`
+	SwapFee        float64 `json:"swapFee"`
```
guiltylotus, line 7: If you define `swapFee` in the config. It means that this `swapFee` is static and applies to all pools. 

But following your [pair contract ](https://bscscan.com/address/0xEDBe30433528643D1EA4733E89D3fFa99BC304B2#readContract), I think swapFee is different for each pool, so I suggest you has an extra struct, write the logic to get fee in `pool_tracker.go`, and save it into `p.Extra `

```
type Extra struct {
	Fee *big.Int `json:"fee"`
}
```

pkg/source/usdfi/type.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+type StaticExtra struct {
+	Stable bool `json:"stable"`
+}
+
+type Reserves struct {
+	Reserve0           *big.Int
+	Reserve1           *big.Int
+	BlockTimestampLast uint32
```
guiltylotus, line 30: Following the ABI, your BlockTimestampLast should be `*big.Int`.
```
{
    "inputs": [
      
    ],
    "name": "getReserves",
    "outputs": [
      {
        "internalType": "uint256",
        "name": "_reserve0",
        "type": "uint256"
      },
      {
        "internalType": "uint256",
        "name": "_reserve1",
        "type": "uint256"
      },
      {
        "internalType": "uint256",
        "name": "_blockTimestampLast",
        "type": "uint256"
      }
    ],
```

pkg/valueobject/exchange.go
```diff
@@ -170,6 +171,7 @@ var AMMSourceSet = map[Exchange]struct{}{
 	ExchangeKyberSwapStatic:     {},
 	ExchangeVelodrome:           {},
 	ExchangeDystopia:            {},
+	ExchangeUsdfi: 				 {},
```
guiltylotus, line 174: Can you keep the same spacing with other lines? It looks more pretty. 🙇‍♂️

pkg/valueobject/exchange.go
```diff
@@ -87,6 +87,7 @@ var (
 
 	ExchangeVelodrome Exchange = "velodrome"
 	ExchangeDystopia  Exchange = "dystopia"
+	ExchangeUsdfi  Exchange = "usdfi"
```
guiltylotus, line 90: Can you keep the same spacing with other lines? It looks more pretty. 🙇‍♂️

guiltylotus:
Overall, You are doing good at getting new pools and pool states. But you are missing `pool_simulator.go`.

`pool_simulator.go` uses to calculate the amount of token B from the amount of token A of a swap. You can see `pool_simulator.go` of the velodrome folder as an example.

pkg/source/usdfi/pool_simulator.go
```diff
@@ -0,0 +1,125 @@
+package velodrome
```
guiltylotus, line 1: should be package `usdfi`

PR #69, opened 2023-07-14
fix: zipswap fail pool

lehainam-dev:


lehainam-dev:
#69. Nice

pkg/source/uniswap/pools_list_updater.go
```diff
@@ -84,7 +85,7 @@ func (d *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte
 		}, []interface{}{&pairAddresses[j]})
 	}
 
-	if _, err := getPairAddressRequest.Aggregate(); err != nil {
+	if _, err := getPairAddressRequest.TryAggregate(); err != nil {
```
it4rb, line 88: if we use `TryAggregate`, maybe we should check `resp.Result` to see which call is failing and ignore it instead of relying on zeroAddress?

it4rb (approved):

pkg/source/uniswap/pools_list_updater.go
```diff
@@ -121,6 +120,10 @@ func (d *PoolsListUpdater) processBatch(ctx context.Context, pairAddresses []com
 	rpcRequest.SetContext(ctx)
 
 	for i := 0; i < limit; i++ {
+		if eth.IsZeroAddress(pairAddresses[i]) {
```
it4rb, line 123: if we choose to skip here, should we ignore at the loop below as well?

pkg/source/uniswap/pools_list_updater.go
lehainam-dev: I updated using `resp.Result` to ignore failed pools.

pkg/source/uniswap/pools_list_updater.go
lehainam-dev: Updated. Since `processBatch` now only processes succeed pairs, I remove this skip logic.

it4rb (approved):

hanguyenk (approved):

PR #78, opened 2023-07-25
refactor: add necessary methods for pool

piavgh:

### DIFF
--- a/pkg/entity/pool.go
+++ b/pkg/entity/pool.go
@@ -25,6 +25,22 @@ type PoolToken struct {
 
 type PoolTokens []*PoolToken
 
+func ClonePoolTokens(poolTokens []*PoolToken) []*PoolToken {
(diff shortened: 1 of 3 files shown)


pkg/entity/pool.go
```diff
@@ -25,6 +25,22 @@ type PoolToken struct {
 
 type PoolTokens []*PoolToken
 
+func ClonePoolTokens(poolTokens []*PoolToken) []*PoolToken {
```
it4rb, line 28: um just curious, why do we need to move this up here?

it4rb (approved):

pkg/entity/pool.go
piavgh: I want to keep the method near its type so it's easier to reason about its purpose and make it hard to misunderstand the method

datluongductuan (approved):
LGTM

PR #82, opened 2023-07-27
maverick: not update binPos of mergeID not equals 0

guiltylotus:

### DIFF
--- a/pkg/source/maverickv1/pool_tracker.go
+++ b/pkg/source/maverickv1/pool_tracker.go
@@ -129,13 +129,12 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
 		}
 		bins[strI] = bin
 
-		if binPositions[bin.LowerTick.String()] == nil {
-			binPositions[bin.LowerTick.String()] = make(map[string]*big.Int)
-		}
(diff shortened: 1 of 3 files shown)


hanguyenk (approved):

pkg/source/maverickv1/pool_tracker.go
```diff
@@ -129,13 +129,12 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
...
 		bins[strI] = bin
 
-		if binPositions[bin.LowerTick.String()] == nil {
-			binPositions[bin.LowerTick.String()] = make(map[string]*big.Int)
-		}
-		binPositions[bin.LowerTick.String()][bin.Kind.String()] = big.NewInt(int64(i))
-
 		if bin.MergeID.Int64() == 0 {
 			d.putTypeAtTick(binMap, bin.Kind, bin.LowerTick)
+			if binPositions[bin.LowerTick.String()] == nil {
```
piavgh, line 134: can you please explain why do we need to move this piece of code inside the `if`?

guiltylotus: the `mergeID != 0` means the bin has been merged, so we don't count that bin. 

Bin at mergeID = 0  includes the reserves of itself and other bins which have mergeID != 0

PR #94, opened 2023-08-03
[AG-612] algebrav1 fee approximation

it4rb:

### DIFF
--- a/pkg/source/algebrav1/data_storage.go
+++ b/pkg/source/algebrav1/data_storage.go
@@ -1,0 +1,482 @@
+package algebrav1
+
+import (
+	"errors"
+	"math/big"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+	"github.com/KyberNetwork/logger"
+)
+
+// port https://github.com/cryptoalgebra/AlgebraV1/blob/dfebf532a27803dafcbf2ba49724740bd6220505/src/core/contracts/libraries/DataStorage.sol
+
+// we won't need full 65536 points as in SC, so use a custom struct
+type TimepointStorage struct {
+	data    map[uint16]Timepoint // original data from SC
+	updates map[uint16]Timepoint // new data written during simulation (or precalculation), will be cleared
+}
+
+func (s *TimepointStorage) Get(index uint16) Timepoint {
+	if v, ok := s.updates[index]; ok {
+		logger.Debugf("access new %v %v", index, v)
+		return v
+	}
+	if v, ok := s.data[index]; ok {
+		logger.Debugf("access exists %v %v", index, v)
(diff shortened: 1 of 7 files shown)


pkg/source/algebrav1/pool_tracker.go
```diff
@@ -188,9 +190,261 @@ func (d *PoolTracker) fetchRPCData(ctx context.Context, p entity.Pool) (FetchRPC
...
 	return res, err
 }
 
+func (d *PoolTracker) approximateFee(ctx context.Context, poolAddress, dataStorageOperator string, state *GlobalState, currentLiquidity *big.Int) error {
+	// fee approximation: assume that the swap will be soon after this
+	blockTimestamp := uint32(time.Now().Unix())
+	yesterday := blockTimestamp - WINDOW
+	timepoints, err := d.getPoolTimepoints(ctx, state.TimepointIndex, poolAddress, yesterday)
+	if err != nil {
+		return err
```
piavgh, line 207: we should add an error log here to detect the problem easier

pkg/source/algebrav1/pool_tracker.go
```diff
@@ -188,9 +190,261 @@ func (d *PoolTracker) fetchRPCData(ctx context.Context, p entity.Pool) (FetchRPC
...
+	}
+
+	if timepoints == nil {
+		// not initialized pool has been locked already, but set here just for sure
+		state.Unlocked = false
```
piavgh, line 212: it's better to add an info log here (to answer the question: why the route does not go through this pool?)

it4rb: I've added log in the parent function, to catch the case when it's locked at the SC

pkg/source/algebrav1/data_storage.go
```diff
@@ -0,0 +1,482 @@
...
+		BlockTimestamp:                0,
+		TickCumulative:                0,
+		SecondsPerLiquidityCumulative: new(big.Int),
+		VolatilityCumulative:          new(big.Int),
+		AverageTick:                   0,
+		VolumePerLiquidityCumulative:  new(big.Int),
+	}
+}
+func (s *TimepointStorage) Set(index uint16, v Timepoint) {
+	s.updates[index] = v
```
piavgh, line 40: why don't we need to set the s.data here?

it4rb: `s.data` is the original data fetched from SC
`s.updates` will contains the written data during simulation or fee approximation, and will be reverted after each calculation (this code is originally for simulator, but then got moved to tracker for fee approximation)
In the Get function we'll check from both `s.data` and `s.updates`

piavgh (approved):

PR #110, opened 2023-08-11
add error catching and unit test for calcInvariant

datluongductuan:

### DIFF
--- a/pkg/source/balancer-composable-stable/pool_simulator.go
+++ b/pkg/source/balancer-composable-stable/pool_simulator.go
@@ -5,8 +5,6 @@ import (
 	"math/big"
 	"strings"
 
-	"github.com/KyberNetwork/logger"
-
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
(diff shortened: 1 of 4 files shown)


pkg/source/balancer-composable-stable/pool_simulator.go
```diff
@@ -192,11 +205,14 @@ func (c *PoolSimulator) _swapWithBptGivenIn(
 			preJoinExitSupply,
 			preJoinExitInvariant,
 		)
+		if err != nil {
```
it4rb, line 208: `_doJoinSwap` might return `nil, nil, nil, nil`, in that case `amountCalculated=nil` so `DivDownFixed`  below might be crashed
please check in the original SC code to see what do they do in that case and update here

it4rb (approved):
one minor comment about a case that might cause crash
but that will be catch at upper layer so it's ok to ignore if fixing that require too much effort

pkg/source/balancer-composable-stable/pool_simulator.go
datluongductuan: The original SC _doJoinSwap is here https://github.com/balancer/balancer-v2-monorepo/blob/872342e060bfc31c3ab6a1deb7b1d3050ea7e19d/pkg/pool-stable/contracts/ComposableStablePool.sol#L386

Since the `CalcAmountOut` in our `poolSimulator` is only used for extract given amountIn (for amountIn, find amountOut), but not used for the other case (for amountOut, find amountIn). So I skip `_joinSwapExactBptOutForTokenIn` implementation and return `nil,nil,nil,nil`.

However, I have added a check to return `ErrorInvalidAmountOutCalculated` if `amountCalculated == nil`

hanguyenk (approved):

PR #113, opened 2023-08-17
AG-645: fix bug invalid swap

guiltylotus:

### DIFF
--- a/pkg/source/maverickv1/math.go
+++ b/pkg/source/maverickv1/math.go
@@ -66,6 +66,7 @@ func swapTick(delta *Delta, state *MaverickPoolState) (*Delta, error) {
 	var currentBins []Bin
 	var err error
 
+	oldActiveTick := new(big.Int).Set(activeTick)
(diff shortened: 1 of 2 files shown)


pkg/source/maverickv1/math.go
```diff
@@ -1100,7 +1116,7 @@ func scaleFromAmount(amount *big.Int, decimals uint8) (*big.Int, error) {
 	}
 }
 
-func scaleToAmount(amount *big.Int, decimals uint8) (*big.Int, error) {
+func ScaleToAmount(amount *big.Int, decimals uint8) (*big.Int, error) {
```
lehainam-dev, line 1119: Why does this function change to public? I don't see anywhere using it outside of the current package.

guiltylotus: Yes, I used it for pool-service CalculateTVL maverick tl.

hanguyenk (approved):

PR #115, opened 2023-08-17
Add bestPaths entity for pregen service

datluongductuan:

### DIFF
--- a/pkg/entity/bestpath.go
+++ b/pkg/entity/bestpath.go
@@ -1,0 +1,29 @@
+package entity
+
+import "encoding/json"
+
+const BestPathKey = "bestpaths"
+
+type BestPath struct {
+	// Pools list address pools that path swap through, length of pools = length of tokens - 1
+	Pools []string
+	// Tokens list tokens that path swap through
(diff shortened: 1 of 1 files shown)


pkg/entity/bestpath.go
```diff
@@ -0,0 +1,29 @@
+package entity
+
+import "encoding/json"
+
+const BestPathKey = "bestpaths"
+
+type BestPath struct {
```
hanguyenk, line 7: Can we reuse `SimpleSwap` struct which is defined here? [Link](https://github.com/KyberNetwork/router-service/blob/main/internal/pkg/valueobject/simple_route.go#L4-L17)
--
In my opinion, we should name it `SimplePath` or `MinimalPath` which indicates that this struct contains minimal data of a path.

datluongductuan: imo I think we should not reuse SimpleSwap struct, since this way will lead to change some other code, make it more complicated at the moment. So currently I will rename to MinimalPath

hanguyenk (approved):

PR #116, opened 2023-08-23
[AG-704][AG-705] update LO pool list updater/tracker multi SCs

it4rb:
Now each limit-order pool need to have info about SC address that was used to create all orders in that pool

### DIFF
--- a/pkg/source/limitorder/pool_tracker.go
+++ b/pkg/source/limitorder/pool_tracker.go
@@ -39,9 +39,15 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
 		token0, token1 = p.Tokens[1], p.Tokens[0]
 	}
 
-	var staticExtra StaticExtra
-	if err := json.Unmarshal([]byte(p.StaticExtra), &staticExtra); err != nil {
-		return entity.Pool{}, err
+	var contractAddress string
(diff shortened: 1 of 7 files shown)


pkg/source/limitorder/pool_tracker.go
```diff
@@ -29,24 +29,36 @@ func NewPoolTracker(cfg *Config) *PoolTracker {
...
 	if strings.ToLower(token0.Address) < strings.ToLower(token1.Address) {
 		token0, token1 = p.Tokens[1], p.Tokens[0]
 	}
 
+	var contractAddress string
+	if d.config.SupportMultiSCs {
+		var staticExtra StaticExtra
+		if err := json.Unmarshal([]byte(p.StaticExtra), &staticExtra); err != nil {
+			return entity.Pool{}, err
+		}
```
lehainam-dev, line 47: So with old orders without p.StaticExtra, I think this line will raise error, and cause the pool will not update, right? Leading we will need to flush all orders after release?

it4rb: yes we would need to remove all old pools anyway, because the pool ID will be changed so the old pools is no longer valid

lehainam-dev (approved):
LGTM

piavgh (approved):

PR #119, opened 2023-08-24
feat: add mapping token address for balancer

lehainam-dev:
related: AG-775

## Why did we need it?

Encoding Balancer data on L2 requires assetIndex (instead of assetAddress). I added this logic, so when building route, L2 encoder has the needed data.


it4rb (approved):

it4rb:
minor: GetMeta will only be called if the route goes through the pool, so less frequent than NewPoolSimulator
maybe we can choose to build the mapping from pool's tokens at GetMeta instead of at NewPoolSimulator so that we don't have to store the map in memory (but I think the memory for that is small anyway so not a big deal)

lehainam-dev:
@it4rb I updated the PR with your suggestion. Thank you, really nice observation.

it4rb (approved):

piavgh (approved):

PR #132, opened 2023-08-30
feat: integrate Kyber PMM

piavgh:
## Why did we need it?

Add pools list updater, pool tracker and pool simulator for Kyber PMM
Encoding logic should be handle in a separate PR

## Related Issue

- https://team-kyber.atlassian.net/browse/AG-753
- https://team-kyber.atlassian.net/browse/AG-754
- https://team-kyber.atlassian.net/browse/AG-755

## How Has This Been Tested?

At the moment, just use the unit tests inside

### DIFF
--- a/pkg/source/kyber-pmm/pool_simulator.go
+++ b/pkg/source/kyber-pmm/pool_simulator.go
@@ -143,10 +143,10 @@ func (p *PoolSimulator) swapBaseToQuote(tokenAmountIn pool.TokenAmount, tokenOut
 		Fee:            &pool.TokenAmount{Token: tokenAmountIn.Token, Amount: integer.Zero()},
 		Gas:            p.gas.Swap,
 		SwapInfo: SwapInfo{
-			BaseToken:        tokenAmountIn.Token,
-			BaseTokenAmount:  tokenAmountIn.Amount.String(),
-			QuoteToken:       tokenOut,
-			QuoteTokenAmount: amountOut.String(),
+			TakerAsset:   tokenAmountIn.Token,
+			TakingAmount: tokenAmountIn.Amount.String(),
+			MakerAsset:   tokenOut,
+			MakingAmount: amountOut.String(),
 		},
 	}, nil
 }
@@ -172,10 +172,10 @@ func (p *PoolSimulator) swapQuoteToBase(tokenAmountIn pool.TokenAmount, tokenOut
 		Fee:            &pool.TokenAmount{Token: tokenAmountIn.Token, Amount: integer.Zero()},
 		Gas:            p.gas.Swap,
 		SwapInfo: SwapInfo{
-			BaseToken:        tokenAmountIn.Token,
-			BaseTokenAmount:  tokenAmountIn.Amount.String(),
-			QuoteToken:       tokenOut,
-			QuoteTokenAmount: amountOut.String(),
+			TakerAsset:   tokenAmountIn.Token,
+			TakingAmount: tokenAmountIn.Amount.String(),
+			MakerAsset:   tokenOut,
+			MakingAmount: amountOut.String(),
 		},
 	}, nil
 }
--- a/pkg/source/kyber-pmm/pool_simulator.go
+++ b/pkg/source/kyber-pmm/pool_simulator.go
@@ -1,0 +1,254 @@
+package kyberpmm
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+	"strings"
+
+	"github.com/KyberNetwork/blockchain-toolkit/float"
+	"github.com/KyberNetwork/blockchain-toolkit/integer"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+)
+
+type PoolSimulator struct {
+	pool.Pool
+	baseToken              *entity.PoolToken
+	quoteToken             *entity.PoolToken
+	baseToQuotePriceLevels []PriceLevel
+	quoteToBasePriceLevels []PriceLevel
+	gas                    Gas
+}
+
+func NewPoolSimulator(entityPool entity.Pool) (*PoolSimulator, error) {
+	var numTokens = len(entityPool.Tokens)
+	var tokens = make([]string, numTokens)
+	var reserves = make([]*big.Int, numTokens)
+
+	if numTokens != 2 {
+		return nil, fmt.Errorf("pool's number of tokens should equal 2")
+	}
+
+	var staticExtra StaticExtra
+	if err := json.Unmarshal([]byte(entityPool.StaticExtra), &staticExtra); err != nil {
+		return nil, err
+	}
+
+	var baseToken, quoteToken *entity.PoolToken
+	for i := 0; i < numTokens; i += 1 {
+		tokens[i] = entityPool.Tokens[i].Address
+		reserves[i] = bignumber.NewBig10(entityPool.Reserves[i])
+
+		if strings.EqualFold(staticExtra.BaseTokenAddress, entityPool.Tokens[i].Address) {
+			baseToken = entityPool.Tokens[i]
+		}
+
+		if strings.EqualFold(staticExtra.QuoteTokenAddress, entityPool.Tokens[i].Address) {
+			quoteToken = entityPool.Tokens[i]
+		}
+	}
+
+	var extra Extra
+	if err := json.Unmarshal([]byte(entityPool.Extra), &extra); err != nil {
+		return nil, err
+	}
+
+	return &PoolSimulator{
+		Pool: pool.Pool{
+			Info: pool.PoolInfo{
+				Address:    strings.ToLower(entityPool.Address),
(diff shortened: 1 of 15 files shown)


pkg/source/kyber-pmm/pools_list_updater.go
```diff
@@ -0,0 +1,133 @@
...
+	return pools, metadataBytes, nil
+}
+
+func (u *PoolsListUpdater) extractPMMPairs(
+	tokens map[string]TokenItem,
+	pmmPairs map[string]PairItem,
+) []entity.Pool {
+	result := make([]entity.Pool, 0, len(pmmPairs))
+	for pairID, pmmPair := range pmmPairs {
+		newPool, err := u.transformToPool(tokens, pairID, pmmPair)
```
it4rb, line 62: should we check if newPool's ID already existed in `result`? (in case they return pair for both A/B and B/A)

it4rb: hmm it seems the `prices` api is returning both bid and ask, so it's likely that they will always return A/B or B/A, not both, so maybe we don't have to check duplicated pool ID
but I'm not very sure about this so please double check with the team again

pkg/source/kyber-pmm/pool_tracker.go
```diff
@@ -0,0 +1,144 @@
...
+			},
+		)
+	}
+
+	for _, ask := range priceLevels.Asks {
+		baseToQuoteAskPrice, err := strconv.ParseFloat(ask[0], 64)
+		if err != nil {
+			continue
+		}
+		quoteToBaseBidPrice := 1 / baseToQuoteAskPrice
```
it4rb, line 126: maybe need to check for 1/0 here, just to be safe

pkg/source/kyber-pmm/pool_simulator.go
```diff
@@ -0,0 +1,254 @@
...
+
+		amountOut = new(big.Float).Add(
+			amountOut,
+			new(big.Float).Mul(
+				swappableAmount, new(big.Float).SetFloat64(priceLevels[currentLevelIdx].Price),
+			),
+		)
+
+		amountInLeft = new(big.Float).Sub(amountInLeft, swappableAmount)
+		currentLevelIdx += 1
```
it4rb, line 208: is the returned priceLevel always sorted from best price to worst?
in paraswap doc they said:
> it is expected that the entries start with the best price and get progressively worse (however this is not enforced, so any price curve can be used)

maybe we need to confirm with trading team about this

pkg/source/kyber-pmm/pool_tracker.go
it4rb: another question: is it enough to use float64, or should we use decimal instead? I don't know if there is any case where the price is big enough for the float to be imprecise

pkg/source/kyber-pmm/pool_simulator.go
```diff
@@ -0,0 +1,254 @@
...
+	).Int(nil)
+
+	return &pool.CalcAmountOutResult{
+		TokenAmountOut: &pool.TokenAmount{Token: tokenOut, Amount: amountOut},
+		Fee:            &pool.TokenAmount{Token: tokenAmountIn.Token, Amount: integer.Zero()},
+		Gas:            p.gas.Swap,
+		SwapInfo: SwapInfo{
+			BaseToken:        tokenAmountIn.Token,
+			BaseTokenAmount:  tokenAmountIn.Amount.String(),
+			QuoteToken:       tokenOut,
```
it4rb, line 177: in this case QuoteToken should be tokenIn, right?

pkg/source/kyber-pmm/pool_simulator.go
```diff
@@ -0,0 +1,254 @@
...
+			amountOut,
+			new(big.Float).Mul(
+				swappableAmount, new(big.Float).SetFloat64(priceLevels[currentLevelIdx].Price),
+			),
+		)
+
+		amountInLeft = new(big.Float).Sub(amountInLeft, swappableAmount)
+		currentLevelIdx += 1
+
+		if amountInLeft.Cmp(float.Zero()) == 0 || currentLevelIdx > len(priceLevels)-1 {
```
it4rb, line 210: is there any case where this float comparison here lead to incorrect result?
how about we transform raw amount to raw*decimal amount at pool-tracker, and only use integer amount in pool-simulator?

pkg/source/kyber-pmm/pools_list_updater.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func NewPoolsListUpdater(cfg Config, client IClient) *PoolsListUpdater {
+	return &PoolsListUpdater{
+		config: cfg,
+		client: client,
+	}
+}
+
+func (u *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte) ([]entity.Pool, []byte, error) {
```
lehainam-dev, line 26: Do we have cases when a pair is removed from the API?

lehainam-dev, line 26: Can the PPM pairs be removed from API?

pkg/source/kyber-pmm/pool_simulator.go
piavgh: no, it does not guarantee that the levels are sorted by price

We just have to go through the price levels step by step (because the response of price levels is an array, so the order will be maintained)

it4rb: hmm so I think maybe we should sort it at pool-tracker, so that we can have maximum amountOut here?

pkg/source/kyber-pmm/pool_simulator.go
piavgh: I thought about that, but we are not able to do it

For example, this is the price levels of KNC/USDT from the API https://staging-kyberswap.knreserve.com/kyberswap/v1/prices:

```
"KNC/USDT": {
      "bids": [
        [
          "0.4062988393151176",
          "98.4499489770304"
        ]
      ],
      "asks": [
        [
          "0.531859877600558",
          "25.000000443678086"
        ]
      ]
    },
```

It means up to 98.4499489770304 KNC can be sold at the price of 0.4062988393151176 USDT / 1 KNC

The decimals of USDT is 6, so if we convert `0.4062988393151176` to USDT decimals, we can not get `*big.Int`

That's why I have to use *big.Float in this case

it4rb: um what I meant is to multiply "98.4499489770304" with KNC's decimals, so that we don't have to divide in sim

piavgh: We will have to multiply it with the price anyway https://github.com/KyberNetwork/kyberswap-dex-lib/pull/132/files#diff-ddca6ce3c783c6250d72fa03c8d2fd349a9d970f47e0730dd5a456e9a81a8d21R202-R204

So if we store the amount with decimals (using *big.Float in this case), we will have also to store the price with decimals.
Personally, I want to keep the pool state as what is provided from the source (transforming to *big.Float then marshal it to JSON to store is a bit unnecessary)

pkg/source/kyber-pmm/pool_simulator.go
piavgh: Good catch, I updated the code to make it easier to understand
Now I'm using takerAsset and makerAsset

pkg/source/kyber-pmm/pool_simulator.go
piavgh: no, I confirmed with the foundation trading team, that we MUST respect the price levels order

The reason they said this is because it's the incentive for the maker (PMM team), maybe they only want to give a better price when it reaches a specified amount (for example, you can only have better price if you bought at least 1000 shares)

it4rb (approved):

pkg/source/kyber-pmm/pools_list_updater.go
piavgh: yes, they will only return A/B or B/A but not both
In the pool-service, we also check if poolId exists before saving it to the data store, so I think in this function, it's not necessary to check it

pkg/source/kyber-pmm/pool_tracker.go
piavgh: good catch, I added a check to avoid division by 0 panic

About your second question, because even the PMM team provides the rate in float64 (not decimals), so there will be no problem

pkg/source/kyber-pmm/pools_list_updater.go
lehainam-dev: What I'm concerned about is if a pair is removed from API, we will have to delete it/update its pool's reserve to 0.

piavgh: yes, but the PMM API supports us already

https://developers.paraswap.network/augustusrfq-api-specification#prices-endpoint

> Note that if you wish to remove a pair from trading, it should not be removed from /pairs endpoint straight away.  Instead, it must be retained there and given prices of {} as shown above for WETH/USDT for a period of time, after which it can finally be removed.
Failing to respond to this endpoint or responding with an error will cause market making to be temporarily disabled for this market maker, which may be done intentionally if you wish to cease trading entirely for a period of time.

So in the pool tracker, if the pair is to be removed, it's state will be empty anyway

lehainam-dev (approved):

PR #136, opened 2023-09-08
fetch limit-order Operator signature

it4rb:
- add RFQ handler to fetch Operator signature (will be called by router-service when encoding route)
- allow to filter by SC address
- add FeeConfig


pkg/source/limitorder/pools_list_updater.go
```diff
@@ -57,6 +64,10 @@ func (d *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte
 func (d *PoolsListUpdater) extractTokenPairs(loPairs []*limitOrderPair) []*tokenPair {
 	pairMap := make(map[string]*tokenPair, 0)
 	for _, loPair := range loPairs {
+		// if we're filtering by address, and if this pool SC doesn't match, then ignore
+		if d.includedContractAddresses.Cardinality() > 0 && !d.includedContractAddresses.Contains(strings.ToLower(loPair.ContractAddress)) {
```
lehainam-dev, line 68: Minor: I think only the second-half condition is enough for this check.

hanguyenk (approved):

pkg/source/limitorder/pools_list_updater.go
it4rb: the first condition is in case we deploy new code but haven't update the config for prod yet, then `includedContractAddresses` will be empty and we won't ignore pool so it will still works as before

lehainam-dev: Oh right, I remember the TDD/understand now. My bad. Thank you for the explanation.

PR #140, opened 2023-09-11
fix: handle error insufficient liquidity in PMM price levels

piavgh:
## Why did we need it?

If the amount in is greater than the available amount that price levels can provide, we should return error insufficient liquidity instead of allowing the simulation to continue


hanguyenk (approved):

lehainam-dev (approved):
LGTM

PR #147, opened 2023-09-18
feat: integrate iziswap

lehainam-dev:
related: AG-856

Mostly forked from https://github.com/opcc22059/kyberswap-dex-lib/tree/iZiSwap

NOTE: PoolListUpdater calls external IZumiSwap API.


pkg/source/iziswap/math.go
```diff
@@ -0,0 +1,8 @@
+package iziswap
+
+import "math/big"
+
+func NewBig10(s string) (res *big.Int) {
```
hanguyenk, line 5: I think we can use `blockchain-tool-kit` instead of having this method

pkg/source/iziswap/pool_list_query.go
```diff
@@ -0,0 +1,87 @@
...
+	Data  []PoolInfo `json:"data,omitempty"`
+	Total int        `json:"total"`
+}
+
+func getPoolsList(
+	ctx context.Context,
+	client *http.Client,
+	params *PoolsListQueryParams,
+) ([]PoolInfo, error) {
+	req, err := http.NewRequest(http.MethodGet, urlRoot, nil)
```
hanguyenk, line 48: I think the whole logic of this function should be grouped in a repository/adapter struct. You can refer limit order / kyber pmm

pkg/source/iziswap/math.go
lehainam-dev: I refactored this into manual parse to big.Int.

pkg/source/iziswap/pool_list_query.go
lehainam-dev: Updated to separate client.

pkg/source/iziswap/pool_simulator.go
```diff
@@ -0,0 +1,180 @@
...
+	}, nil
+}
+
+func (p *PoolSimulator) CalcAmountOut(
+	tokenAmountIn pool.TokenAmount,
+	tokenOut string,
+) (*pool.CalcAmountOutResult, error) {
+
+	var tokenInIndex = p.GetTokenIndex(tokenAmountIn.Token)
+	var tokenOutIndex = p.GetTokenIndex(tokenOut)
```
hanguyenk, line 77: I think we don't need to iterate through the list of tokens twice like this. How about
```go
if tokenAmountIn.Token == p.Tokens[0] && tokenOut == p.Tokens[1]  || tokenAmountIn.Token == p.Tokens[1] && tokenOut == p.Tokens[0] {
  # do something
}

# do something else
```

pkg/source/iziswap/pool_simulator.go
```diff
@@ -0,0 +1,180 @@
...
+	if !ok {
+		logger.Warn("failed to UpdateBalance for UniV3 pool, wrong swapInfo type")
+		return
+	}
+	p.PoolInfo.CurrentPoint = si.nextPoint
+	p.PoolInfo.Liquidity = si.nextLiquidity
+	p.PoolInfo.LiquidityX = si.nextLiquidityX
+}
+
+func (p *PoolSimulator) GetMetaInfo(tokenIn string, tokenOut string) interface{} {
```
hanguyenk, line 174: What is the purpose of this method?

lehainam-dev: I return the LimitPoint since this data is needed for encoding.

pkg/source/iziswap/pool_simulator.go
lehainam-dev: I think using `GetTokenIndex` is more readable, don't you think?

I also updated the source, to using early return to reduce nest level.

hanguyenk (approved):

PR #148, opened 2023-09-18
fix (curve-meta): add underlaying tokens in get tokens

lehainam-dev:
## Why did we need it?

Context: https://team-kyber.slack.com/archives/C04R9NSNEKF/p1695017808835559
Summary: While `curve-meta` can swap to underlying tokens, the current `GetTokens()` function only returns the direct tokens in pools. This leads to the wrong calculate `MinHopsToTokenOut`, due to the wrong create `tokenToPoolAddress` ([code](https://github.com/KyberNetwork/router-service/blob/main/internal/pkg/usecase/findroute/spfav2/best_route.go#L32-L37)).

This PR fixes the above issue.

### DIFF
--- a/pkg/source/curve/meta/pool_simulator.go
+++ b/pkg/source/curve/meta/pool_simulator.go
@@ -261,6 +261,13 @@ func (t *Pool) GetMetaInfo(tokenIn string, tokenOut string) interface{} {
 	}
 }
 
+func (t *Pool) GetTokens() []string {
+	var result []string
+	result = append(result, t.GetInfo().Tokens...)
+	result = append(result, t.BasePool.GetInfo().Tokens...)
+	return result
+}
+
 func (t *Pool) getUnderlyingIndex(token string) int {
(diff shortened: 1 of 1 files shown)


pkg/source/curve/meta/pool_simulator.go
```diff
@@ -261,6 +261,13 @@ func (t *Pool) GetMetaInfo(tokenIn string, tokenOut string) interface{} {
 	}
 }
 
+func (t *Pool) GetTokens() []string {
+	var result []string
```
hanguyenk, line 265: might be a shorter way
```go
return append(t.GetInfo().Tokens, t.BasePool.GetInfo().Tokens...)
```

lehainam-dev: I'm scared it can modify the `t.GetInfo().Tokens` 😨 , so I created a new var to be safe.
Is it necessary?

it4rb (approved):

hanguyenk (approved):

PR #151, opened 2023-09-19
Implement a new dex handler for uniswapv3

hiepnv90:
Implement a new dex handler for uniswapv3

## Why did we need it?

Update PoolTracker to use pool-ticks service for fetching pool ticks data, pool-ticks service is an event driven service for keeping pool ticks up to date compare to on-chain data. The reason for implement a new dex handler is that we can keep the old code unchanged and unaffected.


bbenzikry:
Hi - Sorry to barge in on your PR flow. 
Given the separation done here, I was wondering whether you are going to release different services the library interacts with as OSS as well.

hiepnv90:
@bbenzikry We are considering to do it.

bbenzikry:
@hiepnv90 Great. I appreciate the quick response :)

PR #154, opened 2023-09-21
Feat/update interface for pool tracker

hiepnv90:
Update IPoolTracker interface for accepting more params

## Why did we need it?

We need to pass event logs for pool to handle properly:
- We need to fetch pool state from node at a specified block number
- We can use event logs to update uniswapv2 reserves instead of fetching it from node to reduce cost

### DIFF
--- a/pkg/source/algebrav1/pool_tracker.go
+++ b/pkg/source/algebrav1/pool_tracker.go
@@ -15,7 +15,7 @@ import (
 	"github.com/sourcegraph/conc/pool"
 
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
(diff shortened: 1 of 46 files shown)


pkg/source/algebrav1/pool_tracker.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/sourcegraph/conc/pool"
 
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+	dexpool "github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
```
piavgh, line 18: `sourcePool` is better.

We have some liquidity sources that are not dex (limit order, kyber PMM, hashflow, maybe more in the future)

piavgh (approved):

PR #169, opened 2023-10-02
Update hasReserves()

guiltylotus:

### DIFF
--- a/pkg/source/wombat/pool_tracker.go
+++ b/pkg/source/wombat/pool_tracker.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/KyberNetwork/ethrpc"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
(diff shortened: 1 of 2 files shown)


pkg/source/wombat/pool_tracker.go
```diff
@@ -143,6 +143,11 @@ func (d *PoolTracker) GetNewPoolState(ctx context.Context, p entity.Pool) (entit
 			}
 		}
 
+		// This pool has token in subgraph but not in contract: https://bscscan.com/address/0x2ea772346486972e7690219c190dadda40ac5da4#readProxyContract
+		if strings.EqualFold(assetAddresses[i].Hex(), EmptyAddress) {
```
hanguyenk, line 147: You can use [ZeroAddress](https://github.com/KyberNetwork/blockchain-toolkit/blob/main/account/constants.go#L14)  from blockchain-toolkit

guiltylotus: Updated

hanguyenk (approved):

PR #186, opened 2023-10-12
AG-875: update univ2 state from logs

lehainam-dev:


pkg/source/uniswap/pool_tracker.go
```diff
@@ -22,24 +23,46 @@ func NewPoolTracker(
...
-		ABI:    uniswapV2PairABI,
-		Target: p.Address,
-		Method: pairMethodGetReserves,
-		Params: nil,
-	}, []interface{}{&reserves})
+	latestSyncEvent := findLatestSyncEvent(params.Logs)
+	if latestSyncEvent == nil {
+		logger.WithFields(logger.Fields{
+			"poolAddress": p.Address,
+		}).Debug("Fetch reserves from node")
```
hiepnv90, line 42: I think we can change this log to info because it's unlikely to happen, we can use it to see how frequent we have to fetch data from node.

hiepnv90 (approved):

pkg/source/uniswap/pool_tracker.go
lehainam-dev: Updated. Thank you.

hanguyenk (approved):

PR #201, opened 2023-10-25
feat: support pol-matic on ethereum

hanguyenk:
- There is only one pool for this liquidity source. It's PolygonMigration contract.
- There is no pool state as the pool will always return 1:1 rate
- Token0 will always have to be Matic to determine whether the swap is `migrate` or `unmigrate`. If the tokenIn is matic, it's migrate, otherwise, it's unmigrate.


pkg/source/pol-matic/pool_list_updater.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		newMetadataBytes = []byte(`{"hasInitialized": true}`)
+	}
+
+	// Token0 has to be Matic, otherwise it will break swap logic
+	return []entity.Pool{
+		{
+			Address: strings.ToLower(u.config.PolygonMigrationAddress),
+			Tokens: []*entity.PoolToken{
+				{Address: strings.ToLower(matic.String()), Decimals: maticDecimals, Swappable: true},
```
it4rb, line 154: minor: maybe we should check if maticDecimals==polygonDecimals just for sure (our sim logic rely on that)

it4rb (approved):

pkg/source/pol-matic/pool_list_updater.go
hanguyenk: The contract assumes both pol and matic have the same decimals. It simply transfers the exact amountIn of the other token.
image

lehainam-dev (approved):

PR #212, opened 2023-11-02
Integrate smardex

ChauNguyen-K:
Integrate smardex to our system.

## Why did we need it?
- Smardex smart contract: https://github.com/SmarDex-Dev/smart-contracts/blob/main/contracts/core/SmardexPair.sol

## Related Issue
https://team-kyber.atlassian.net/browse/AG-1108

## How Has This Been Tested?
- Unit test base on input from pr: https://github.com/KyberNetwork/kyberswap-dex-lib/pull/196/files


pkg/source/smardex/pool_list_tracker.go
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	if u.config.ChainID == 1 {
+		pairFee = PairFee{
+			FeesPool: big.NewInt(2),
+			FeesLP:   big.NewInt(5),
+			FeesBase: FEES_BASE_ETHEREUM,
+		}
+	} else {
+		pairFee = PairFee{
+			FeesBase: FEES_BASE_ETHEREUM,
```
lehainam-dev, line 74: I think this should be `FEES_BASE`, as other chains have `uint256 public constant FEES_BASE = 1_000_000;`. 
```suggestion
			FeesBase: FEES_BASE,
```

pkg/source/smardex/pool_list_tracker.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if err != nil {
+		logger.WithFields(logger.Fields{
+			"poolAddress": p.Address,
+			"error":       err,
+			"dex":         u.config.DexID}).Errorf(
+			"failed to marshal pool extra")
+		return entity.Pool{}, err
+	}
+
```
lehainam-dev, line 153: We may also want to update the latest timestamp, which indicates the latest timestamp that we have tracked in this pool. I think we should add:
```suggestion
	p.Timestamp = time.Now().Unix()
```

pkg/source/smardex/pool_list_tracker.go
```diff
@@ -0,0 +1,160 @@
...
+			FeesPool: big.NewInt(2),
+			FeesLP:   big.NewInt(5),
+			FeesBase: FEES_BASE_ETHEREUM,
+		}
+	} else {
+		pairFee = PairFee{
+			FeesBase: FEES_BASE_ETHEREUM,
+		}
+		rpcRequest.AddCall(&ethrpc.Call{
+			ABI:    pairABI,
```
lehainam-dev, line 77: As you point out, the SC on Ethereum may differ from other chains, so maybe using the same `pairABI` for all chains might not work? (the method `getPairFees` is not defined in the `abi/Pair.json`)?

pkg/source/smardex/pool_list_tracker.go
lehainam-dev, line 1: Minor: Other integrations name this file as `pool_tracker.go`. I think it is better to keep it consistent with others. Do you agree?

pkg/source/smardex/pool_list_updater_test.go
```diff
@@ -0,0 +1,83 @@
...
+		PoolPagingSize: 20,
+		ChainID:        uint(1),
+	}
+
+	ts.updater = PoolListUpdater{
+		config:       &config,
+		ethrpcClient: ts.client,
+	}
+
+}
```
lehainam-dev, line 41: Umm, for me it is weird to call other services when running tests (in this case call to our internal RPC), since external services can always fail, leading the test can be flaky; and since tests also run on our CI, it will repeatedly call our RPC. 

Mocking RPC responses is not good either, since it adds too much work.

My suggestion is that we can skip this test (still keep the test file, just add `t.Skip()`). When we want to run it (mostly on local), we can commend out the skip flag.

Wdyt? I'd love to hear your thoughts on this.

pkg/source/smardex/pool_list_tracker.go
ChauNguyen-K: It seems correct, 10000 is outdated value, and feeLP and feePool are changed too, pr: https://github.com/SmarDex-Dev/smart-contracts/pull/7/files. I will fix it, thanks.

pkg/source/smardex/pool_list_tracker.go
ChauNguyen-K: Pair.json contains getPairFees too, it is just an interface used to pack and unpack response from sc. I tested and it works.

pkg/source/smardex/pool_list_tracker.go
ChauNguyen-K: updated file name, thanks.

pkg/source/smardex/pool_list_tracker.go
lehainam-dev: Umm, in the `pool_list_tracker_test.go`, if I update the config to Base network, it throws an error `failed to build call data for target=0x9fcF8F5Bd54db123470c96620441cA5C342A8Bd4 method=getPairFees, err: method 'getPairFees' not found`.

Here is what I updated in the test:
- Change RPC endpoint to `https://base.kyberengineering.io`.
- Change Multicall contract to `0xcA11bde05977b3631167028862bE2a173976CA11`.
- Change FactoryAddress to `0xdd4536dD9636564D891c919416880a3e250f975A`.
- Change ChainID to `8453`.

pkg/source/smardex/pool_list_updater_test.go
ChauNguyen-K: I agree, I intended to skip the tests too, however when I explore ethrpc repo, we still enable some tests like this, then I think it's ok. But, we usually update ethrpc, so skip these tests in dex-lib is better, thank you.

pkg/source/smardex/pool_list_tracker.go
lehainam-dev: I ran with Base network config and it throws error `method=getPairFees, err: method 'getPairFees' not found`.

ChauNguyen-K: Sorry, I confused your comments with getFees method, I have just updated the code, tested on bsc, polygon and arbitrum too.

pkg/source/smardex/pool_list_tracker.go
lehainam-dev: Um, now I'm confused with your new changes.
I'm sorry if I'm not clear in my previous review, however, I think your previous `pairFee` with Ethereum is correct, only `pairFee` on other chains need to be updated.

For reference:
1. [Here is a Smardex Pair on Ethereum](https://etherscan.io/address/0x9fcF8F5Bd54db123470c96620441cA5C342A8Bd4#code): As defined in `SmardexLibrary.sol`, feesLP = 5, feesPool = 2, feesBase = 10000 = your previous FEES_BASE_ETHEREUM. Everything is good.
2. [Here is a Smardex Pair on Polygon](https://polygonscan.com/address/0x77476148B72ECB4E9f6A3cDDC5dd437Df1F003F3#code) which define in `SmardexLibrary.sol` `FEES_BASE = 1_000_000`, however in the old code, you set this feesBase = FEES_BASE_ETHEREUM (=10000), which I suggested to update to `FEES_BASE` (=1000000) instead.

pkg/source/smardex/math.go
```diff
@@ -0,0 +1,96 @@
...
+ */
+func approxEq(x *big.Int, y *big.Int) bool {
+	res := big.NewInt(0)
+	if x.Cmp(y) == 1 {
+		return x.Cmp(res.Mul(y, APPROX_EQ_PRECISION).Div(res, APPROX_EQ_BASE_PRECISION).Add(res, y)) == -1
+	}
+	return y.Cmp(res.Mul(x, APPROX_EQ_PRECISION).Div(res, APPROX_EQ_BASE_PRECISION).Add(res, x)) == -1
+}
+
+func sqrt(value *big.Int) *big.Int {
```
hanguyenk, line 38: I think we can move these functions to `blockchain-toolkit` to reuse. I did with dsmath: 
https://github.com/KyberNetwork/blockchain-toolkit/blob/main/dsmath/dsmath.go#L15-L23

pkg/source/smardex/pool_simulator.go
```diff
@@ -0,0 +1,180 @@
...
+	}, nil
+
+}
+
+func (t *PoolSimulator) GetMetaInfo(_ string, _ string) interface{} {
+	return nil
+}
+
+func (p *PoolSimulator) UpdateBalance(params poolpkg.UpdateBalanceParams) {
+	si, ok := params.SwapInfo.(SwapInfo)
```
hanguyenk, line 170: I'm not sure if this implementation could potentially introduce any bugs, as I haven't previously implemented `UpdateBalance` in this way (although it does seem quite innovative). 

Previously, we didn't have the SwapInfo data, which we recently added with through `kyberswap-limit-order`. Because of this, we often had to manually recalculate updated values, in this case are `reserves` and `fictive reserves`.

ChauNguyen-K: The smardex formula has side-affect due to timestamp, if we re-calculate we will get the new price average out which differs from the previous value we used to calculate the path. I see that we have many pools implement this way, example: https://github.com/KyberNetwork/kyberswap-dex-lib/blob/7a231f222b7d9875e000a53dbf7f2ad6800b0e90/pkg/source/maverickv1/pool_simulator.go#L139 and https://github.com/KyberNetwork/kyberswap-dex-lib/blob/7a231f222b7d9875e000a53dbf7f2ad6800b0e90/pkg/source/pancakev3/pool_simulator.go#L194

pkg/source/smardex/math.go
ChauNguyen-K: Fixed

pkg/source/smardex/pool_simulator.go
```diff
@@ -0,0 +1,191 @@
...
+		zeroForOne = true
+	}
+
+	var (
+		fictiveReserveIn   *big.Int = p.FictiveReserve.FictiveReserve0
+		fictiveReserveOut  *big.Int = p.FictiveReserve.FictiveReserve1
+		priceAverageIn     *big.Int = p.PriceAverage.PriceAverage0
+		priceAverageOut    *big.Int = p.PriceAverage.PriceAverage1
+		balanceIn          *big.Int = new(big.Int).Sub(p.GetReserves()[0], p.FeeToAmount.Fees0)
+		balanceOut         *big.Int = new(big.Int).Sub(p.GetReserves()[1], p.FeeToAmount.Fees1)
```
lehainam-dev, line 73: Since the `getReserves()` function already calculate:
```
        reserve0_ = IERC20(token0).balanceOf(address(this)) - feeToAmount0;
        reserve1_ = IERC20(token1).balanceOf(address(this)) - feeToAmount1;
```

I think these lines should be changed to:
```suggestion
		balanceIn          *big.Int = p.GetReserves()[0]
		balanceOut         *big.Int = p.GetReserves()[1]
```

pkg/source/smardex/pool_simulator.go
```diff
@@ -0,0 +1,191 @@
...
+		userTradeTimestamp          = now().Unix()
+	)
+	if !zeroForOne {
+		fictiveReserveIn = p.FictiveReserve.FictiveReserve1
+		fictiveReserveOut = p.FictiveReserve.FictiveReserve0
+		priceAverageIn = p.PriceAverage.PriceAverage1
+		priceAverageOut = p.PriceAverage.PriceAverage0
+		balanceIn = new(big.Int).Sub(p.GetReserves()[1], p.FeeToAmount.Fees1)
+		balanceOut = new(big.Int).Sub(p.GetReserves()[0], p.FeeToAmount.Fees0)
```
lehainam-dev, line 82: Same with the above comment.

pkg/source/smardex/smardex_library.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if param.priceAverageIn.Cmp(zero) != 1 || param.priceAverageOut.Cmp(zero) != 1 {
+		return result, ErrInsufficientPriceAverage
+	}
+
+	feesTotalReversed := new(big.Int)
+	feesTotalReversed.Sub(param.feesBase, param.feesLP).Sub(feesTotalReversed, param.feesPool)
+	amountInWithFees := new(big.Int)
+	amountInWithFees.Mul(param.amount, feesTotalReversed).Div(amountInWithFees, param.feesBase)
+	firstAmountIn := computeFirstTradeQtyIn(param)
```
lehainam-dev, line 73: The `amountInWithFees` should be passed into `computeFirstTradeQtyIn`.

Reference from SC (I check in this [Polygon pair](https://polygonscan.com/address/0x77476148B72ECB4E9f6A3cDDC5dd437Df1F003F3#code)):
```
        uint256 _firstAmountIn = computeFirstTradeQtyIn(
            SmardexLibrary.GetAmountParameters({
                amount: _amountInWithFees,
        ...
```

pkg/source/smardex/pool_simulator.go
ChauNguyen-K: Fixed, thank you.

lehainam-dev (approved):
LGTM

hanguyenk (approved):

PR #216, opened 2023-11-08
feat: implement new handler for uniswap-v2

hanguyenk:
## Why did we need it?
- Add `blockNumber` to pool data
- Reimplement PoolSimulator
- Add test cases
- Refactor code


pkg/source/uniswap-v2/pool_list_updater.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// listPairAddresses lists address of pairs from offset
+func (u *PoolsListUpdater) listPairAddresses(ctx context.Context, offset int, batchSize int) ([]common.Address, error) {
+	listPairAddressesResult := make([]common.Address, batchSize)
+
+	listPairAddressesRequest := u.ethrpcClient.NewRequest().SetContext(ctx)
+
+	for i := 0; i < batchSize; i++ {
+		index := big.NewInt(int64(offset + i + 1))
```
lehainam-dev, line 149: Should not add `+ 1` here. It causes the first pool will never be fetched.

pkg/source/uniswap-v2/constant.go
```diff
@@ -0,0 +1,20 @@
+package uniswapv2
+
+const (
+	DexType = "uniswap-v2"
```
lehainam-dev, line 4: Minor: Should we follow with the current `uniswapv3` convention?

lehainam-dev: It also matches with Golang package naming convention.

pkg/source/uniswap-v2/pool_simulator_test.go
```diff
@@ -0,0 +1,163 @@
...
+						Reserves: []*big.Int{utils.NewBig("10089138480746"), utils.NewBig("10066716097576")},
+					},
+				},
+				fee:          utils.NewBig("3"),
+				feePrecision: utils.NewBig("1000"),
+			},
+			params: poolpkg.UpdateBalanceParams{
+				TokenAmountIn:  poolpkg.TokenAmount{Token: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48", Amount: utils.NewBig("125224746")},
+				TokenAmountOut: poolpkg.TokenAmount{Token: "0xdac17f958d2ee523a2206206994597c13d831ec7", Amount: utils.NewBig("124570062")},
+				Fee:            poolpkg.TokenAmount{Token: "0x3041cbd36888becc7bbcbc0045e3b1f144466f5f", Amount: utils.NewBig("375674")},
```
lehainam-dev, line 102: Minor: should the Fee token be TokenIn `0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48`?

pkg/source/uniswap-v2/pool_list_updater.go
lehainam-dev: Updated.

PR #224, opened 2023-11-14
add SwapLimit interface

cyberliem:
## Why did we need it?

https://www.notion.so/kybernetwork/8ff731d012614053b5f1918995837a32?v=5cfc182035ad4dc0997657fd6b532c9c&p=18faa39e551941f38960071794539cae&pm=s


pkg/source/pool/swap_limit.go
```diff
@@ -0,0 +1,8 @@
+package pool
+
+import "math/big"
+
+type SwapLimit interface {
+	GetLimit(key string) *big.Int
```
piavgh, line 6: we should add a comment to explain clearly how this interface and it's functions will do

pkg/liquidity-source/uniswap-v2/pool_simulator.go
```diff
@@ -39,6 +39,10 @@ type (
 	}
 )
 
+func (p *PoolSimulator) CalculateLimit() map[string]*big.Int {
```
piavgh, line 42: to reduce code duplication, I think we can move this code into the `pool/pool.go` (because PoolSimulator embeds the Pool struct inside)

For PMM and Synthetix, we can define a custom function for CalculateLimit (it will shadow the default CalculateLimit function in `pool/pool.go`

piavgh: similar to this CanSwapFrom function: https://github.com/KyberNetwork/kyberswap-dex-lib/blob/d2bcc708dc317ddbf8c18a1317495ba60ec73ab5/pkg/source/pool/pool.go#L48-L52

piavgh (approved):

PR #239, opened 2023-11-20
improve: calcAmountOut maverick-v1

lehainam-dev:

### DIFF
--- a/pkg/source/maverickv1/math.go
+++ b/pkg/source/maverickv1/math.go
@@ -482,89 +482,89 @@ func tickPrice(tickSpacing *big.Int, activeTick *big.Int) (*big.Int, error) {
 	}
 
 	var ratio *big.Int
-	if new(big.Int).And(tick, bignumber.NewBig("0x1")).Cmp(zeroBI) != 0 {
+	if tick.Bit(0) != 0 {
 		ratio = bignumber.NewBig("0xfffcb933bd6fad9d3af5f0b9f25db4d6")
 	} else {
 		ratio = bignumber.NewBig("0x100000000000000000000000000000000")
 	}
 
-	if new(big.Int).And(tick, bignumber.NewBig("0x2")).Cmp(zeroBI) != 0 {
+	if tick.Bit(1) != 0 {
 		ratio.Mul(ratio, bignumber.NewBig("0xfff97272373d41fd789c8cb37ffcaa1c"))
 		ratio.Rsh(ratio, 128)
 	}
-	if new(big.Int).And(tick, bignumber.NewBig("0x4")).Cmp(zeroBI) != 0 {
+	if tick.Bit(2) != 0 {
 		ratio.Mul(ratio, bignumber.NewBig("0xfff2e50f5f656ac9229c67059486f389"))
 		ratio.Rsh(ratio, 128)
 	}
-	if new(big.Int).And(tick, bignumber.NewBig("0x8")).Cmp(zeroBI) != 0 {
+	if tick.Bit(3) != 0 {
 		ratio.Mul(ratio, bignumber.NewBig("0xffe5caca7e10e81259b3cddc7a064941"))
 		ratio.Rsh(ratio, 128)
 	}
-	if new(big.Int).And(tick, bignumber.NewBig("0x10")).Cmp(zeroBI) != 0 {
+	if tick.Bit(4) != 0 {
 		ratio.Mul(ratio, bignumber.NewBig("0xffcb9843d60f67b19e8887e0bd251eb7"))
 		ratio.Rsh(ratio, 128)
 	}
-	if new(big.Int).And(tick, bignumber.NewBig("0x20")).Cmp(zeroBI) != 0 {
+	if tick.Bit(5) != 0 {
 		ratio.Mul(ratio, bignumber.NewBig("0xff973b41fa98cd2e57b660be99eb2c4a"))
 		ratio.Rsh(ratio, 128)
 	}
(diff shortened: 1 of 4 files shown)


pkg/source/maverickv1/math.go
```diff
@@ -921,19 +922,15 @@ func msb(x *big.Int) *big.Int {
 // ------------- maverick basic math --------------------
 func mulDiv(a, b, c *big.Int, ceil bool) (*big.Int, error) {
 	product := new(big.Int).Mul(a, b)
-	if a.Cmp(zeroBI) == 0 || new(big.Int).Div(product, a).Cmp(b) == 0 {
```
lehainam-dev, line 924: This condition seems always true to me. Since `product = a * b`, then `product / a = b`.
As A. Hank explains, Smart contract needs to check this condition for int overflow, since SC uses int256. However, in our simulation, we use big.Int instead, so no overflow case.

lehainam-dev: So, we can remove this condition, or update our code to match the overflow case as the SC.

pkg/source/maverickv1/math.go
```diff
@@ -691,7 +692,7 @@ func getTickL(
 	precisionBump := big.NewInt(0)
 	big40 := big.NewInt(40)
 
-	if new(big.Int).Rsh(reserveA, 60).Cmp(zeroBI) == 0 && new(big.Int).Rsh(reserveB, 60).Cmp(zeroBI) == 0 {
+	if reserveA.Bit(60) == 0 && reserveB.Bit(60) == 0 {
```
it4rb, line 695: um if reserveA.Bit[61]=1 and reserveA.Bit[60]=0 then your check returns true but the original returns false? (I'm not sure if that will happen though)

pkg/source/quickperps/pool_simulator_test.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 func TestPool_CalcAmountOut(t *testing.T) {
+	t.Skip("Skip failing test for now")
```
it4rb, line 15: um why does quickperps test failing here?

lehainam-dev: @it4rb main branch has failed test cases. I skip it now to not affect this PR.

pkg/source/maverickv1/math.go
```diff
@@ -1035,38 +1032,26 @@ func mul(a, b *big.Int) (*big.Int, error) {
...
 }
 
 func sMulDownFixed(a, b *big.Int) (*big.Int, error) {
 	var product = new(big.Int).Mul(a, b)
-	if a.Cmp(zeroBI) == 0 || new(big.Int).Div(product, a).Cmp(b) == 0 {
-		return new(big.Int).Div(product, One), nil
-	}
-
-	return nil, ErrMulOverflow
+	return new(big.Int).Div(product, One), nil
```
it4rb, line 1054: um what's the different between `product` and `new(big.Int).Div(product, One)` (I'm not sure what's the meaning of sMulDownFixed)

pkg/source/maverickv1/math.go
lehainam-dev: You're right. Thank you for catching this.
I will reverse to the old condition.

pkg/source/maverickv1/math.go
lehainam-dev: `One` variable here equals 10^18. It's a bad name, starting from Paraswap's integration.

pkg/source/quickperps/pool_simulator_test.go
lehainam-dev: Will rebase main branch and remove this line after Minh fixes the test.

pkg/source/maverickv1/math.go
lehainam-dev: I only realize this after changing to `return product, nil` and the test fails 🤣 .

pkg/source/maverickv1/math.go
```diff
@@ -1035,38 +1032,26 @@ func mul(a, b *big.Int) (*big.Int, error) {
 
 func mulUpFixed(a, b *big.Int) (*big.Int, error) {
 	product := new(big.Int).Mul(a, b)
+	isNegative := product.Sign() == -1
```
it4rb, line 1035: I'm not sure if we'll ever get negative result here (SC will because of overflow), but all these stuff are quite tricky so maybe it's better to keep it

guiltylotus (approved):

guiltylotus:
LGTM!

it4rb (approved):

hanguyenk (approved):

pkg/source/maverickv1/math.go
lehainam-dev: I understand what you're concerned about, but still a little hesitant to reverse it.

These changes increase ~300 rounds in benchmark (~1.28%).

Since the condition `a.Cmp(zeroBI) == 0 || new(big.Int).Div(product, a).Cmp(b) == 0` seems always true in Golang, I kind of want to remove it.

I think we either remove the condition (as I did) or add new code to handle int256 overflow case (which is more complicated). I kind of don't want to keep the current code.

Also, I can't judge if it is worth the 1% improvement.

What do you think?

it4rb: sorry for not being clear, I meant maybe we don't need to check `isNegative` because it might never happen with bigInt (this func always get non-negative input IIRC), but anw it's too risky to remove that

PR #246, opened 2023-11-28
Fix quickperps testcase

guiltylotus:

### DIFF
--- a/pkg/source/quickperps/pool_simulator_test.go
+++ b/pkg/source/quickperps/pool_simulator_test.go
@@ -1,8 +1,10 @@
 package quickperps
 
 import (
+	"encoding/json"
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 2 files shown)


pkg/source/quickperps/pool_simulator_test.go
```diff
@@ -71,13 +73,24 @@ func TestPool_CalcAmountOut(t *testing.T) {
...
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			var extra Extra
+			err := json.Unmarshal([]byte(tc.entityPool.Extra), &extra)
+			assert.Nil(t, err)
+
+			extra.Vault.PriceFeed.PriceFeedProxies["0x4f9a0e7fd2bf6067db6994cf12e4495df938e6e9"].Timestamp = uint32(time.Now().Unix())
+			extra.Vault.PriceFeed.PriceFeedProxies["0xc5015b9d9161dca7e18e32f6f25c4ad850731fd4"].Timestamp = uint32(time.Now().Unix())
```
lehainam-dev, line 88: The timestamp values might change in the future.
Can the test still pass if the timestamp changes?

Maybe instead of setting `Timestamp = uint32(time.Now().Unix())`, should we hardcoded the timestamp value instead?

guiltylotus:
```
timestamp := big.NewInt(int64(priceFeed.Timestamp))
if new(big.Int).Add(timestamp, pf.ExpireTimeForPriceFeed).Cmp(big.NewInt(time.Now().Unix())) <= 0 {
   return nil, ErrVaultPriceFeedExpired
}
```

This condition checks `timestamp + pf.ExpireTimeForPriceFeed must > time.Now()` so when I set timestamp = time.Now(). Expect this check always pass

lehainam-dev (approved):

hanguyenk (approved):

PR #250, opened 2023-12-04
wombat: update balance

guiltylotus:

### DIFF
--- a/pkg/source/wombat/wombatmain/math.go
+++ b/pkg/source/wombat/wombatmain/math.go
@@ -45,7 +45,7 @@ func Swap(
 		return nil, nil, err
 	}
 
-	newToAssetCash := removeCash(toAsset.Cash, new(big.Int).Add(actualToAmount, haircut))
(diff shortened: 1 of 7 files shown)


pkg/source/wombat/wombatmain/math.go
```diff
@@ -45,7 +45,7 @@ func Swap(
 		return nil, nil, err
 	}
 
-	newToAssetCash := removeCash(toAsset.Cash, new(big.Int).Add(actualToAmount, haircut))
+	newToAssetCash := new(big.Int).Add(toAsset.Cash, new(big.Int).Add(actualToAmount, haircut))
```
lehainam-dev, line 48: Uhmm you called `removeCash` before, but now updated to `new(big.Int).Add`?
Just want to double check this.

guiltylotus: I have a question here https://team-kyber.slack.com/archives/C04SNUEPC4S/p1701679258079639/

I removed removeCash, and addCash function

hanguyenk (approved):

PR #263, opened 2023-12-13
refactor: balancer v1 v2 + velocore v2

vuquang23:

### DIFF
--- a/pkg/liquidity-source/balancer-v1/pool_simulator.go
+++ b/pkg/liquidity-source/balancer-v1/pool_simulator.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"math/big"
 
+	"github.com/KyberNetwork/blockchain-toolkit/number"
 	"github.com/KyberNetwork/logger"
 	"github.com/goccy/go-json"
 	"github.com/holiman/uint256"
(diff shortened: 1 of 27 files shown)


pkg/liquidity-source/balancer-v1/pool_simulator.go
```diff
@@ -214,3 +239,25 @@ func (s *PoolSimulator) swapExactAmountIn(
 
 	return tokenAmountOut, spotPriceAfter, nil
 }
+
+func (s *PoolSimulator) validateAmountIn(tokenIn string, amountIn *uint256.Int) error {
+	bMulBalanceInAndMaxIn, err := BNum.BMul(s.records[tokenIn].Balance, BConst.MAX_IN_RATIO)
```
hanguyenk, line 244: why don't we just use `s.maxTotalAmountsIn[tokenIn]`?

pkg/liquidity-source/balancer-v2/composable-stable/type.go
```diff
@@ -55,8 +56,8 @@ type StaticExtra struct {
 	PoolType       string         `json:"poolType"`
 	PoolTypeVer    int            `json:"poolTypeVer"`
 	BptIndex       int            `json:"bptIndex"`
-	VaultAddress   string         `json:"vaultAddress"`
 	ScalingFactors []*uint256.Int `json:"scalingFactors"`
+	Vault          string         `json:"vault"`
```
hanguyenk, line 60: [small] I think we don't need to change `VaultAddress` to `Vault` but it's fine to keep

hanguyenk (approved):

PR #264, opened 2023-12-14
support CalcAmountIn interface

it4rb:

### DIFF
--- a/pkg/source/pool/iface.go
+++ b/pkg/source/pool/iface.go
@@ -50,7 +50,7 @@ type RFQResult struct {
 	Extra        any
 }
 
-type IPoolReverseSimulator interface {
+type IPoolExactOutSimulator interface {
 	// CalcAmountIn calculate the `amountIn` of `tokenIn` needed to get `tokenAmountOut`
 	// caller might need to run `CalcAmountOut` again to determine if the returned `amountIn` is good enough
 	CalcAmountIn(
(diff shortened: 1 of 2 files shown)


pkg/source/pool/iface.go
```diff
@@ -50,6 +50,15 @@ type RFQResult struct {
 	Extra        any
 }
 
+type IPoolReverseSimulator interface {
```
hanguyenk, line 53: There are 2 trade types: exact in and exact out. I think we can name following the trade type: exactOutSimulator

pkg/source/pool/iface.go
```diff
@@ -50,6 +50,15 @@ type RFQResult struct {
...
+	// CalcAmountIn calculate the `amountIn` of `tokenIn` needed to get `tokenAmountOut`
+	// caller might need to run `CalcAmountOut` again to determine if the returned `amountIn` is good enough
+	CalcAmountIn(
```
hanguyenk, line 56: Should we keep the interface similar to `CalcAmountOut`?

it4rb: um the result might be not be exact (the returned `amountIn` might not yield enough `amountOut`), so the caller might need to run CalcAmountOut again anyway to verify, I think they can get other param like `gas` from that 2nd call

pkg/source/pool/iface.go
it4rb, line 53:
```suggestion
type IPoolExactOutSimulator interface {
```

hanguyenk (approved):

PR #266, opened 2023-12-15
feat: integrate ramses v2

lehainam-dev:
## Why did we need it?

Previous ramses-v2 implementation has errors (feeTier & tickSpacing), causing some pools cannot route through.

This PR fixes that issue. This ramses-v2 simulation still heavily bases on uniswapv3.


lehainam-dev:
I keep the integration without new features: pool ticks worker & cross tick gas estimation (to reduce complexity of the PR).

hanguyenk (approved):

PR #268, opened 2023-12-18
add ApproxAmountIn method

it4rb:

### DIFF
--- a/pkg/source/pool/pool.go
+++ b/pkg/source/pool/pool.go
@@ -2,6 +2,7 @@ package pool
 
 import (
 	"errors"
+	"fmt"
 	"math/big"
 
 	"github.com/KyberNetwork/logger"
@@ -9,6 +10,8 @@ import (
 
 var (
 	ErrCalcAmountOutPanic = errors.New("calcAmountOut was panic")
+	ErrInsufficientAmount = errors.New("not enough amount")
+	ErrNotConverge        = errors.New("secant loop cannot converged")
 )
 
 type Pool struct {
@@ -133,6 +136,19 @@ type CalcAmountOutParams struct {
 	Limit         SwapLimit
 }
 
+type CalcAmountInParams struct {
+	TokenAmountOut TokenAmount
+	TokenIn        string
+	Limit          SwapLimit
+}
+
+type CalcAmountInResult struct {
+	TokenAmountIn *TokenAmount
+	Fee           *TokenAmount
+	Gas           int64
+	SwapInfo      interface{}
+}
+
 // wrap around pool.CalcAmountOut and catch panic
 func CalcAmountOut(pool IPoolSimulator, tokenAmountIn TokenAmount, tokenOut string, limit SwapLimit) (res *CalcAmountOutResult, err error) {
 	defer func() {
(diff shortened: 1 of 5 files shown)


pkg/source/pool/pool.go
```diff
@@ -153,3 +169,141 @@ func CalcAmountOut(pool IPoolSimulator, tokenAmountIn TokenAmount, tokenOut stri
...
+
+	// otherwise try to approximate:
+	// consider the pool as a function `fpool`
+	// we need to find amountIn such that fpool(amountIn) = amountOut
+	// in another word, we need to find root of function fpool_1 where fpool_1(x) = fpool(x) - amountOut
+	// here we'll use https://en.wikipedia.org/wiki/Secant_method for that
+
+	// get 1st initial point by converting back expectedAmountOut to tokenIn
+	// this might yield error if the pool doesn't support that
+	x0res, err := CalcAmountOut(pool, expectedTokenOut, param.TokenIn, param.Limit)
```
piavgh, line 241: I think we can make it clearer by defining a separate function `findRootSecant` like this (later on we can have other functions such as `findRootNewton` or `findRootBisection`, ...

```
// Replace float64 to *big.Int, I use float64 here to make the equation easier to read
func findRootSecant(f func(float64) float64, x0, x1, eps float64) float64 {
	var (
		x2 = x1 - f(x1)*(x1-x0)/(f(x1)-f(x0))
	)

	for math.Abs(x2-x1) > eps {
        // logic to handle convergence should be added somewhere inside
		x0, x1 = x1, x2
		x2 = x1 - f(x1)*(x1-x0)/(f(x1)-f(x0))
	}

	return x2
}
```

The passed in `f` function can be defined outside:

```
f := func(x *big.Int) *big.Int {
        // f = CalcAmountOut(x) - exactAmountOut
		return new(big.Int).Sub(CalcAmountOut(x), exactAmountOut)
	}
```

pkg/source/pool/pool.go
```diff
@@ -153,3 +169,141 @@ func CalcAmountOut(pool IPoolSimulator, tokenAmountIn TokenAmount, tokenOut stri
...
+				TokenAmountOut: fx1Res.TokenAmountOut,
+				Fee:            fx1Res.Fee,
+				Gas:            fx1Res.Gas,
+				SwapInfo:       fx1Res.SwapInfo,
+			}, nil
+		}
+
+		// if fx0 is close enough to fx1 we should stop to avoid error
+		// (and because we likely won't be able to get any closer to the result)
+		if fx0.Cmp(fx1) == 0 {
```
piavgh, line 291: should we compare to a small epsilon here instead of strictly equal to 0?

it4rb: um this if to avoid setting `denom` to zero and causing error, so I think we should just compare to 0

pkg/source/pool/pool.go
it4rb: I've tried to split it out to another function as suggested, but it's getting quite hairy:
- I would like to re-use 1 call to `CalcAmountOut` to calculate both initial point x1 and fx0Res (before the loop), so the `findRootSecant` func will need to take in 1 extra arg for fx0Res
  - (this is because this approximation will get called a lot when calculating profile, so I would like to call CalcAmountOut as few times as possible)
    - (this is why I cannot call `f(x1)` and `f(x0)` like in your formula above, and had to use previous fx1 to be next fx0)
- in order to be generic, the function `findRootSecant` will need to take in a func from bigInt to bigInt, that mean we'll discard other info (gas, fee...), so after finding the root we'll have to calculate them again, or find another way to keep these info

So I think we can keep it as is for now and refactor later if we need to support multiple find root methods, is that ok?

pkg/source/pool/pool.go
it4rb: ah I see the comment is confusing, at first I wrote this code in `combine-path` library, they use float instead of bigInt, so I have to compare with epsilon and so the comment is "if fx0 is close enough to fx1", but here we're working with bigInt so I think we can compare exact equal

piavgh (approved):

PR #283, opened 2023-12-27
fix dodo negative output (check reserves)

it4rb:


pkg/source/dodo/pool_simulator_test.go
```diff
@@ -108,3 +109,42 @@ func TestCanSwapTo(t *testing.T) {
 func decStr(amt int64) string {
 	return new(big.Int).Mul(big.NewInt(amt), bignumber.TenPowInt(18)).String()
 }
+
+func TestUpdateBalance(t *testing.T) {
```
piavgh, line 113:
```suggestion
func TestCalcAmountOut_PoolDepleted(t *testing.T) {
```

piavgh (approved):

PR #299, opened 2024-01-11
fix bin update for maverick

it4rb:
- Return error if pool doesn't have any bin
- Fix bin update
Paraswap code: https://github.com/paraswap/paraswap-dex-lib/blob/de2a6a2/src/dex/maverick-v1/maverick-math/maverick-pool-math.ts#L231

### DIFF
--- a/pkg/source/maverickv1/math.go
+++ b/pkg/source/maverickv1/math.go
@@ -620,15 +620,19 @@ func adjustAB(bin *Bin, delta *Delta, thisBinAmount, totalAmount, activeTick *bi
 
 	if delta.TokenAIn {
 		bin.ReserveA = new(big.Int).Add(bin.ReserveA, deltaIn)
-		bin.ReserveB = big.NewInt(0)
+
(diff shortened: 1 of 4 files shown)


pkg/source/maverickv1/math.go
```diff
@@ -620,15 +620,19 @@ func adjustAB(bin *Bin, delta *Delta, thisBinAmount, totalAmount, activeTick *bi
 
 	if delta.TokenAIn {
 		bin.ReserveA = new(big.Int).Add(bin.ReserveA, deltaIn)
-		bin.ReserveB = big.NewInt(0)
+
 		if delta.Excess.Cmp(zeroBI) <= 0 {
 			bin.ReserveB = clip(bin.ReserveB, deltaOut)
```
it4rb, line 625: in old code `bin.ReserveB` already set to 0 above, so this will always returns 0

hanguyenk (approved):

PR #300, opened 2024-01-11
[EX-85] - Update dex lib

ChauNguyen-K:
Update dex lib constants to be used in contract-encoding project


pkg/pooltypes/pooltypes.go
```diff
@@ -150,6 +150,10 @@ type Types struct {
 	LegacyBalancerStable           string
 	LegacyBalancerMetaStable       string
 	LegacyBalancerComposableStable string
+	CrowdswapV2                    string
```
lehainam-dev, line 153: Uhm I don't think we have `poolTypes.CrowdswapV2`.
`CrowdswapV2` should be the poolExchange, with poolType of `biswap`.

https://router-api.kyberengineering.io/arbitrum/api/v1/pools?ids=0xa0d518c5f4828a6e6574625301f3a45623138224

This also applies to Aerodrome (velodrome-v2), Thena (velodrome) & PearlV2 (pearl).

You can use the constant variable in each pool folder, rather than define it as a string instead.

ChauNguyen-K: Some pool types are updated redundantly, I removed them.

lehainam-dev (approved):
LGTM

hanguyenk (approved):

PR #306, opened 2024-01-12
add NewPoolSimulatorV2 for uniswapv3

it4rb:
Benchmark:
Before:
`BenchmarkNewPoolSimulator-8   	    2350	    511776 ns/op	  242664 B/op	    3778 allocs/op`

After:
If we don't reuse PoolSimulator instance
`BenchmarkNewPoolSimulatorV2-8   	    8212	    146271 ns/op	  157368 B/op	    2673 allocs/op`

Best case: if we keep reusing the same PoolSimulator (don't need to alloc new tick list array)
`BenchmarkNewPoolSimulatorV2ReuseBestcase-8   	   12393	     95657 ns/op	   87663 B/op	      88 allocs/op`


pkg/source/uniswapv3/pool_simulator.go
```diff
@@ -119,6 +128,97 @@ func NewPoolSimulator(entityPool entity.Pool, chainID valueobject.ChainID) (*Poo
...
+		Liquidity    *big.Int  `json:"liquidity"`
+		SqrtPriceX96 *big.Int  `json:"sqrtPriceX96"`
+		Tick         *big.Int  `json:"tick"`
+		Ticks        []TickGob `json:"ticksGob"`
+	}
+}
+
+func NewPoolSimulatorV2(entityPool entity.Pool, chainID valueobject.ChainID) (*PoolSimulatorV2, error) {
+	p := &PoolSimulatorV2{}
+	err := InitPoolSimulatorV2(entityPool, p, chainID)
```
cyberliem, line 143: looks weird, since we are not pre-loc p. Why do we have to write this type of pattern.

pkg/source/uniswapv3/pool_simulator.go
```diff
@@ -119,6 +128,97 @@ func NewPoolSimulator(entityPool entity.Pool, chainID valueobject.ChainID) (*Poo
...
+		tokens[0] = entityPool.Tokens[0].Address
+		reserves[0] = NewBig10(entityPool.Reserves[0])
+		tokens[1] = entityPool.Tokens[1].Address
+		reserves[1] = NewBig10(entityPool.Reserves[1])
+	}
+
+	// TickGob should be fully compatible with v3Entities.Tick (GobBigInt is just a wrapper around bigInt)
+	// this should always be checked with a unit test
+	// also, uninitialized tick should be ignored at pool-tracker/tick-based-worker already, so no need to check here
+	v3Ticks := *(*v3TickList)(unsafe.Pointer(&p.extra.Ticks))
```
cyberliem, line 176: will this assertion panic if the p.extra.Ticks is empty?

cyberliem (commented):
There are questions

pkg/source/uniswapv3/pool_simulator.go
it4rb: um this is for caller (for example combine-path lib) that want to simply create a simulator without worrying about pre-allocating simulator instance
(router-service will try to call `InitPoolSimulatorV2` directly on the re-used instance)

pkg/source/uniswapv3/pool_simulator.go
it4rb: 've added a testcase for empty list

pkg/source/uniswapv3/type.go
```diff
@@ -59,6 +61,12 @@ type Tick struct {
 	LiquidityNet   *big.Int `json:"liquidityNet"`
 }
 
+type TickGob struct {
+	Index          int                  `json:"index"`
```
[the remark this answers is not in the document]
it4rb: um in this PR we only use Gob for 2 bigInt (liq gross and liq net), not the whole struct, so we'll still need these tags

PR #313, opened 2024-01-16
reuse D and vPrice in curve-meta instead of re-calculating

it4rb:
Issue 1:
Curve-meta is calling `t.BasePool.GetVirtualPrice()` inside `_xp_mem` (https://github.com/KyberNetwork/kyberswap-dex-lib/pull/313/files#diff-9a159db6433983dec62bd91411fbf325a97bd332f85a4be258f0e6024303a39dR10)
But its caller always have `vPrice` calculated already, and the calculation is expensive, so we should pass from the caller instead

Issue 2:
In `GetDyUnderlying`, curve-meta will call `t.BasePool.GetVirtualPrice` then `t.BasePool.GetDy`
`t.BasePool.GetVirtualPrice` has calculated `D` already, and it's expensive, so we should return `D` in `t.BasePool.GetVirtualPrice` then pass it to `t.BasePool.GetDy` to be re-used

Benchmark:
Before
`BenchmarkGetDyUnderlying-8   	   80980	     14481 ns/op	   11297 B/op	     224 allocs/op`

After
`BenchmarkGetDyUnderlying-8   	  141639	      8081 ns/op	    6645 B/op	     136 allocs/op`

Staging:
Before
Screenshot 2024-01-16 at 16 06 35

After
Screenshot 2024-01-16 at 16 06 55

### DIFF
--- a/pkg/source/curve/meta/pool_simulator.go
+++ b/pkg/source/curve/meta/pool_simulator.go
@@ -24,7 +24,7 @@ type ICurveBasePool interface {
 	GetInfo() pool.PoolInfo
 	GetTokenIndex(address string) int
 	// return both vPrice and D
-	GetVirtualPrice() (*big.Int, *big.Int, error)
+	GetVirtualPrice() (vPrice *big.Int, D *big.Int, err error)
 	// if `dCached` is nil then will be recalculated
 	GetDy(i int, j int, dx *big.Int, dCached *big.Int) (*big.Int, *big.Int, error)
 	CalculateTokenAmount(amounts []*big.Int, deposit bool) (*big.Int, error)
--- a/pkg/source/curve/meta/pool_simulator.go
+++ b/pkg/source/curve/meta/pool_simulator.go
@@ -23,8 +23,10 @@ import (
 type ICurveBasePool interface {
 	GetInfo() pool.PoolInfo
 	GetTokenIndex(address string) int
-	GetVirtualPrice() (*big.Int, error)
-	GetDy(i int, j int, dx *big.Int) (*big.Int, *big.Int, error)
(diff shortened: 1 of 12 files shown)


pkg/source/curve/meta/pool_simulator.go
```diff
@@ -23,8 +23,10 @@ import (
 type ICurveBasePool interface {
 	GetInfo() pool.PoolInfo
 	GetTokenIndex(address string) int
-	GetVirtualPrice() (*big.Int, error)
-	GetDy(i int, j int, dx *big.Int) (*big.Int, *big.Int, error)
+	// return both vPrice and D
+	GetVirtualPrice() (*big.Int, *big.Int, error)
```
phqb, line 27: The return data should be named to clarify the second *big.Int is D cached.

phqb (approved):

lehainam-dev (approved):
LGTM. Don't know why the contract implement that way, does it cost them more gas?

it4rb:
@lehainam-dev : 
- for issue1: the contract's `_xp_mem` receives `vp_rate` (virtual price) from the caller instead of calling to base pool (https://github.com/curvefi/curve-contract/blob/d4e8589ac92c4019b3064b2f3a8a87dbc3281b46/contracts/pool-templates/meta/SwapTemplateMeta.vy#L253)

- for issue2: the contract will cache virtual price for 10min (https://github.com/curvefi/curve-contract/blob/d4e8589ac92c4019b3064b2f3a8a87dbc3281b46/contracts/pool-templates/meta/SwapTemplateMeta.vy#L261), so it's less expensive

I'm not really sure why our implementation is different though

PR #315, opened 2024-01-17
[Fix] gmx-gpl race condition (continue)

phqb:

### DIFF
--- a/pkg/source/gmx-glp/pool_simulator_stake.go
+++ b/pkg/source/gmx-glp/pool_simulator_stake.go
@@ -55,7 +55,7 @@ func (p *PoolSimulator) addLiquidityForAccount(swapInfo *gmxGlpSwapInfo, tokenIn
 
 func (p *PoolSimulator) BuyUSDG(swapInfo *gmxGlpSwapInfo, token string, tokenAmount *big.Int) (*big.Int, error) {
 	//_validate(whitelistedTokens[_token], 16);  // canSwapTo vaildated it
-	p.vault.UseSwapPricing = true
+	useSwapPricing := true
 
 	// uint256 tokenAmount = _transferIn(_token);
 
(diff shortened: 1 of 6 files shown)


pkg/source/gmx-glp/pool_simulator_stake.go
```diff
@@ -55,7 +55,7 @@ func (p *PoolSimulator) addLiquidityForAccount(swapInfo *gmxGlpSwapInfo, tokenIn
 
 func (p *PoolSimulator) BuyUSDG(swapInfo *gmxGlpSwapInfo, token string, tokenAmount *big.Int) (*big.Int, error) {
 	//_validate(whitelistedTokens[_token], 16);  // canSwapTo vaildated it
-	p.vault.UseSwapPricing = true
```
it4rb, line 58: please remove this field in vault struct as well

pkg/source/gmx-glp/yearn_token_vault.go
```diff
@@ -52,22 +69,30 @@ func (y *YearnTokenVault) Deposit(amount *big.Int) (*big.Int, error) {
...
-func (y *YearnTokenVault) Withdraw(maxShares *big.Int) (*big.Int, error) {
+func (y *YearnTokenVault) Withdraw(maxShares *big.Int) (*big.Int, *YearnTokenVault, error) {
 	shares := new(big.Int).Set(maxShares)
 	if shares.Cmp(bignumber.ZeroBI) <= 0 {
-		return nil, ErrYearnTokenVaultWithdrawNothing
+		return nil, nil, ErrYearnTokenVaultWithdrawNothing
 	}
 
 	value := y.shareValue(shares)
 	vaultBalance := new(big.Int).Set(y.TotalIdle)
```
it4rb, line 79: will this function get call multiple times within 1 swap? if so then maybe we'll need to check and use `yModified.TotalIdle` here

phqb: Withdraw() is called once at CalcAmountOut(). But I lifted `yModified` out of `Withdraw()` to make `Withdraw()` able to be called multiple times.

pkg/source/gmx-glp/pool_simulator_stake.go
phqb: done

it4rb (approved):

PR #320, opened 2024-01-18
Add tests in pkg/liquidity-sources to check if CalcAmountOut() is concurrent safe

phqb:

### DIFF
--- a/pkg/liquidity-source/velocore-v2/wombat-stable/pool_simulator_test.go
+++ b/pkg/liquidity-source/velocore-v2/wombat-stable/pool_simulator_test.go
@@ -1,0 +1,130 @@
+package wombatstable
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/goccy/go-json"
+	"github.com/stretchr/testify/require"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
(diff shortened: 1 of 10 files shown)


pkg/liquidity-source/velocore-v2/wombat-stable/pool_simulator_test.go
```diff
@@ -0,0 +1,130 @@
...
+// 		return "", err
+// 	}
+
+// 	poolEncoded, _ := json.MarshalIndent(poolEntity, "", "\t")
+
+// 	return string(poolEncoded), nil
+// }
+
+// func TestGetPoolEntity(t *testing.T) {
+// 	poolEncoded, err := getPoolEntity("0x61cb3a0c59825464474ebb287a3e7d2b9b59d093")
```
it4rb, line 127: um just curious, we can get test pool data from redis or from router api (https://router-api.kyberengineering.io/linea/api/v1/pools?ids=0x61cb3a0c59825464474ebb287a3e7d2b9b59d093), is there anything missing there that you have to do this?

it4rb (approved):

pkg/liquidity-source/velocore-v2/wombat-stable/pool_simulator_test.go
phqb: I just hadn't known we have an API to get pool state.

PR #322, opened 2024-01-19
use int/uint256 for univ3

it4rb:
Benchmark:
**CalcAmountOut**
Before:
```
BenchmarkCalcAmountOut-8                  	   43399	     27103 ns/op	   39050 B/op	     787 allocs/op
BenchmarkCalcAmountOutCrossFewTick-8      	   28562	     41896 ns/op	   57493 B/op	    1172 allocs/op
BenchmarkCalcAmountOutCrossManyTick-8     	    4903	    239083 ns/op	  301602 B/op	    6315 allocs/op
```

After:
```
BenchmarkCalcAmountOutV2-8                	  107085	     11221 ns/op	   12192 B/op	     204 allocs/op
BenchmarkCalcAmountOutCrossFewTickV2-8    	   65976	     18309 ns/op	   18596 B/op	     358 allocs/op
BenchmarkCalcAmountOutCrossManyTickV2-8   	   10000	    105416 ns/op	   97454 B/op	    2309 allocs/op
```

BenchmarkCalcAmountOut: doesn't cross tick
BenchmarkCalcAmountOutCrossFewTick: cross 6 ticks
BenchmarkCalcAmountOutCrossManyTick: cross 79 ticks

**NewPoolSimulator** :
Before:
`BenchmarkNewPoolSimulator-8               	    2325	    516261 ns/op	  242604 B/op	    3778 allocs/op`
After
`BenchmarkNewPoolSimulatorV2-8             	    3356	    363253 ns/op	  147010 B/op	    1606 allocs/op`


pkg/source/uniswapv3/type.go
```diff
@@ -59,13 +69,26 @@ type Tick struct {
...
 	Liquidity    *big.Int `json:"liquidity"`
 	SqrtPriceX96 *big.Int `json:"sqrtPriceX96"`
 	Tick         *big.Int `json:"tick"`
 	Ticks        []Tick   `json:"ticks"`
 }
 
+type ExtraTickU256 struct {
+	Liquidity    *uint256.Int `json:"liquidity"`
+	SqrtPriceX96 *uint256.Int `json:"sqrtPriceX96"`
+	Tick         *big.Int     `json:"tick"`
```
hieuvo, line 88: why don't we use TickU256 for `Tick` field 🤔 ?

it4rb: it's a signed integer so we can't use `uint256.Int` (we can use `int256.Int` but this is a single field not an array, so the performance different is not that much, so I decided to keep using big.Int for simplicity)

pkg/source/uniswapv3/pool_simulator_bigint.go
```diff
@@ -0,0 +1,271 @@
...
+	if !ok {
+		logger.Warn("failed to UpdateBalance for UniV3 pool, wrong swapInfo type")
+		return
+	}
+	p.V3Pool.SqrtRatioX96 = si.nextStateSqrtRatioX96
+	p.V3Pool.Liquidity = si.nextStateLiquidity
+	p.V3Pool.TickCurrent = si.nextStateTickCurrent
+}
+
+func (p *PoolSimulatorBigInt) GetMetaInfo(tokenIn string, tokenOut string) interface{} {
```
hieuvo, line 267: minor 
```suggestion
func (p *PoolSimulatorBigInt) GetMetaInfo(_ string, _ string) interface{} {
```

pkg/source/uniswapv3/type.go
hieuvo: minor, for better consistency and readability, if it's the only remaining field using big.Int in a struct, we should convert it to its 256 alternative also

pkg/source/uniswapv3/pool_simulator_bigint.go
it4rb: fixed

pkg/source/uniswapv3/type.go
it4rb: fixed, changed to `int` (it's an `int24` in contract code, so we can use `int` here)

hieuvo (approved):

pkg/source/uniswapv3/type.go
```diff
@@ -59,13 +69,26 @@ type Tick struct {
...
 	Liquidity    *big.Int `json:"liquidity"`
 	SqrtPriceX96 *big.Int `json:"sqrtPriceX96"`
 	Tick         *big.Int `json:"tick"`
 	Ticks        []Tick   `json:"ticks"`
 }
 
+type ExtraTickU256 struct {
+	Liquidity    *uint256.Int `json:"liquidity"`
+	SqrtPriceX96 *uint256.Int `json:"sqrtPriceX96"`
+	Tick         *int         `json:"tick"`
```
lehainam-dev, line 88: Can we change this to just `int`? (remove the pointer?)

it4rb: um this is to check if the `tick` field is missing or not (if we use `int` then missing will result in `tick==0`)

PR #330, opened 2024-01-24
add univ2 InitPoolSimulator to allow caller to re-use sim instance

it4rb:
`InitPoolSimulator` take in an instance of pool simulator and fill it instead of allocating new one

Benchmark:
Allocating new simulator every time:
`BenchmarkNewPoolSimulator-8     	 2421100	       485.7 ns/op	     464 B/op	      13 allocs/op`

Best case: always re-use the same simulator:
`BenchmarkNewPoolSimulatorV2-8   	 8776808	       135.8 ns/op	      48 B/op	       2 allocs/op`

### DIFF
--- a/pkg/liquidity-source/uniswap-v2/pool_simulator.go
+++ b/pkg/liquidity-source/uniswap-v2/pool_simulator.go
@@ -76,6 +76,10 @@ func InitPoolSimulator(entityPool entity.Pool, sim *PoolSimulator) error {
 	if len(entityPool.Tokens) != NUM_TOKEN || len(entityPool.Reserves) != NUM_TOKEN {
 		return errors.New("Invalid number of token")
 	}
+	// in case the caller mess thing up
+	if sim == nil {
+		return errors.New("Invalid simulator instance")
+	}
 
 	var extra Extra
 	if err := json.Unmarshal([]byte(entityPool.Extra), &extra); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/liquidity-source/uniswap-v2/pool_simulator.go
```diff
@@ -61,6 +61,77 @@ func NewPoolSimulator(entityPool entity.Pool) (*PoolSimulator, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	return sim, nil
+}
+
+const NUM_TOKEN = 2
+
+func InitPoolSimulator(entityPool entity.Pool, sim *PoolSimulator) error {
+	if len(entityPool.Tokens) != NUM_TOKEN || len(entityPool.Reserves) != NUM_TOKEN {
```
hieuvo, line 76: just be cautious, we should do nil check with `sim` here

it4rb: Thanks! 've added a check below

pkg/liquidity-source/uniswap-v2/pool_simulator.go
```diff
@@ -61,6 +61,81 @@ func NewPoolSimulator(entityPool entity.Pool) (*PoolSimulator, error) {
...
+	}
+	// in case the caller mess thing up
+	if sim == nil {
+		return errors.New("Invalid simulator instance")
+	}
+
+	var extra Extra
+	if err := json.Unmarshal([]byte(entityPool.Extra), &extra); err != nil {
+		return err
+	}
```
phqb, line 87: Can we reuse `Extra`?

it4rb: yes we could (by putting Extra into the sim struct), but in this case it's quite small so I skipped it (so that I don't have to change NewPoolSimulator v1)

(this case is quite annoying, `extra` only used inside this function and nothing refer to it after that, but somehow the compiler still decided to moved it to heap instead, the heap analysis show the reason is the `json.Unmarshal` call, but I'm still not sure why)

PR #334, opened 2024-01-30
use maverick tick crossed to estimate gas

it4rb:
Each bin crossed will cost some amount of gas, so we should take that into account as well, instead of just using flat constant (similar to univ3)

### DIFF
--- a/pkg/source/maverickv1/math.go
+++ b/pkg/source/maverickv1/math.go
@@ -12,7 +12,7 @@ func GetAmountOut(
 	tokenAIn bool,
 	exactOutput bool,
 	isForPricing bool,
-) (*big.Int, *big.Int, error) {
+) (*big.Int, *big.Int, int, error) {
 	delta := &Delta{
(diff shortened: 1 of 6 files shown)


hanguyenk (approved):

lehainam-dev (approved):
LGTM

pkg/source/maverickv1/math.go
```diff
@@ -34,21 +34,21 @@ func GetAmountOut(
...
+			return nil, nil, 0, err
 		}
 		combine(delta, newDelta)
 
 		// We can not do too much iteration. This variable chosen
 		// as reasonable threshold
 		counter += 1
 		if isForPricing && counter > MaxSwapIterationCalculation {
-			return zeroBI, zeroBI, nil
+			return zeroBI, zeroBI, counter, nil
```
hieuvo, line 45: should be?
```suggestion
			return zeroBI, zeroBI, MaxSwapIterationCalculation, nil
```

PR #365, opened 2024-03-01
fix: uniswap-v2 safe math

piavgh:
## Why did we need it?

Without SafeMath implementation, this pool https://polygonscan.com/address/0x96c7fc08d8cdacdb95a8613b19fffe4d54307263#code still gives a valid amount in = `1` for getAmountin() function with params:

```
reserveIn:        number.NewUint256("1160689189059097452"),
reserveOut:       number.NewUint256("1161607"),
amountOut:        number.NewUint256("500000000"),
```

But in reality, it should revert with the error `ds-math-sub-underflow`

This incorrect behavior breaks the CalcAmountIn function in onchain-price-service

### DIFF
--- a/pkg/liquidity-source/uniswap-v2/pool_simulator.go
+++ b/pkg/liquidity-source/uniswap-v2/pool_simulator.go
@@ -4,12 +4,11 @@ import (
 	"errors"
 	"math/big"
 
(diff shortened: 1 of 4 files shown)


lehainam-dev (approved):
LGTM

PR #366, opened 2024-03-04
track native coin before converting to wrapped

it4rb:


hanguyenk (approved):

pkg/liquidity-source/curve/plain/type.go
```diff
@@ -10,6 +10,9 @@ type (
 		// some Plain pools have oracle for 2nd coin's rate
 		// (not to be confused with Plain2Price that use multiple oracles for all of its coins)
 		Oracle string `json:",omitempty"`
+
+		// which coins are originally native (before being converted to wrapped)
+		IsNativeCoin []bool
```
hieuvo, line 15: for better meaning, it should be in plural form

PR #369, opened 2024-03-04
curve-stable-ng: pool tracker

it4rb:

### DIFF
--- a/pkg/liquidity-source/curve/stable-ng/math.go
+++ b/pkg/liquidity-source/curve/stable-ng/math.go
@@ -1,0 +1,313 @@
+package stableng
+
+import (
+	"math"
+	"time"
+
+	"github.com/KyberNetwork/blockchain-toolkit/number"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/liquidity-source/curve/shared"
+	"github.com/holiman/uint256"
+)
+
+func xpMem(
+	rates []uint256.Int,
+	balances []uint256.Int,
+) []uint256.Int {
+	// try to put `result` in caller's stack (this func will be inlined)
+	var result [shared.MaxTokenCount]uint256.Int
+	count := xpMem_inplace(rates, balances, result[:])
+	return result[:count]
+}
+
+func xpMem_inplace(
+	rates []uint256.Int,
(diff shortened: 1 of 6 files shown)


hanguyenk (approved):

pkg/liquidity-source/curve/stable-ng/math.go
```diff
@@ -0,0 +1,313 @@
...
+		xp[i].Div(number.SafeMul(&rates[i], &balances[i]), Precision)
+	}
+	return numTokens
+}
+
+func (t *PoolSimulator) _A() *uint256.Int {
+	var t1 = t.extra.FutureATime
+	var a1 = t.extra.FutureA
+	var now = time.Now().Unix()
+	if t1 > now {
```
hieuvo, line 38: very minor, to make it more readable, we should apply `return quickly`
```golang
	if t1 <= a {
		return a1
	}
	var t0 = t.extra.InitialATime
	var a0 = t.extra.InitialA
	if a1.Cmp(a0) > 0 {
	....
```

pkg/liquidity-source/curve/stable-ng/math.go
```diff
@@ -0,0 +1,313 @@
...
+		var t0 = t.extra.InitialATime
+		var a0 = t.extra.InitialA
+		if a1.Cmp(a0) > 0 {
+			return number.Add(
+				a0,
+				number.Div(
+					number.Mul(
+						number.Sub(a1, a0),
```
hieuvo, line 46: instead of separate 2 cases, can we just take the absolute of a1-a0?

it4rb: um yes we can, this code has been ported from Solidity once and then get copied over and over again so not much though go into this

the reason it's written like this in Solidity is because of safemath, they can't write something like `abs(a1-a0)` because if `a1 < a0` then `a1-a0` would have raised exception already

we're using uint256.Int which is unsafe by default, so `a1-a0` will underflow and `-(a1-a0)` will be equal to `a0-a1`, so we can use abs here, but this code has been working for a long time and I'm a bit afraid of those tricky under/overflow so I choose to keep the original ported code

PR #370, opened 2024-03-04
curve-stable-ng: pool simulator

it4rb:

### DIFF
--- a/pkg/liquidity-source/curve/stable-ng/math.go
+++ b/pkg/liquidity-source/curve/stable-ng/math.go
@@ -1,0 +1,313 @@
+package stableng
+
+import (
+	"math"
+	"time"
+
+	"github.com/KyberNetwork/blockchain-toolkit/number"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/liquidity-source/curve/shared"
+	"github.com/holiman/uint256"
+)
+
+func xpMem(
+	rates []uint256.Int,
+	balances []uint256.Int,
+) []uint256.Int {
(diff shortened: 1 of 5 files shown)


pkg/liquidity-source/curve/stable-ng/math.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/KyberNetwork/blockchain-toolkit/number"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/liquidity-source/curve/shared"
+	"github.com/holiman/uint256"
+)
+
+func xpMem(
+	rates []uint256.Int,
+	balances []uint256.Int,
+) []uint256.Int {
+	// try to put `result` in caller's stack (this func will be inlined)
```
hieuvo, line 16: do we have any learning material or articles about using array will help to put in caller's stack? I'm curious to take a look

it4rb: um there are 2 separated details here:
- array (fixed length) will usually be put in stack, instead of in heap like slice (dynamic length). Sometimes array might still get put into heap, for example if we return it out of the current function, or if it's too big... Please check out this series (or just part2) https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-stacks-and-pointers.html

- if this function get inlined, then it's just like we copy all its code and paste to the parent caller, so those variables that in its stack will be put in the caller's stack instead. This only works if the function is simple enough to be inlined, so it's quite tricky. Please read more detail here https://words.filippo.io/efficient-go-apis-with-the-inliner/

PR #380, opened 2024-03-08
implemented kyberPMM is the same

cyberliem:
## Why did we need it?

We need a way to tell if the kyberPMM is updated or not (by comparing new state to the old received one)

### DIFF
--- a/pkg/source/kyber-pmm/pool_simulator.go
+++ b/pkg/source/kyber-pmm/pool_simulator.go
@@ -29,6 +29,27 @@ type PoolSimulator struct {
 	timestamp              int64
 }
 
+func samePrice(p1, p2 []PriceLevel) bool {
+	if len(p1) != len(p2) {
+		return false
+	}
+	priceMap := make(map[float64]float64, len(p1))
+	for _, price := range p1 {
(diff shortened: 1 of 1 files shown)


pkg/source/kyber-pmm/pool_simulator.go
```diff
@@ -29,6 +29,27 @@ type PoolSimulator struct {
 	timestamp              int64
 }
 
+func samePrice(p1, p2 []PriceLevel) bool {
+	if len(p1) != len(p2) {
+		return false
+	}
+	priceMap := make(map[float64]float64, len(p1))
```
hieuvo, line 36: be careful as handling float numbers
https://goplay.space/#Po2AoyGVwkl

secmask: @cyberliem levels not need to be ordered, actually we usually give bad price at the first level(due gas fee).
I think compare two price levels by asks vs asks, bids vs bids, and they are slices, so you can do slice compare.
in theory, [{qty: 1, price 0.5}, {qty: 1, price: 0.49}] is different with [{qty: 1, price: 0.49}, {qty: 1, price 0.5}]
because we apply it from top -> end

PR #391, opened 2024-03-15
curve-stable-meta-ng: update plain/stable-ng to become base pool

it4rb:

### DIFF
--- a/pkg/liquidity-source/curve/plain/math.go
+++ b/pkg/liquidity-source/curve/plain/math.go
@@ -311,10 +311,13 @@ func (t *PoolSimulator) getYD(
 	tokenIndex int,
 	xp []uint256.Int,
 	d *uint256.Int,
-) (*uint256.Int, error) {
+
+	//output
+	y *uint256.Int,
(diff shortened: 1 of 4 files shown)


hanguyenk (approved):

pkg/liquidity-source/curve/plain/math.go
```diff
@@ -547,35 +595,44 @@ func (t *PoolSimulator) RemoveLiquidityOneCoin(tokenAmount *big.Int, i int) (*bi
...
-	var err = t.getD(xp, A, &D)
+	var err = t.getD(xp, A, D)
 	if err != nil {
-		return nil, nil, err
+		return err
 	}
 	if t.LpSupply.IsZero() {
-		return nil, nil, ErrDenominatorZero
+		return ErrDenominatorZero
 	}
```
hieuvo, line 635: should we move this if to the top, just to return early if supply is 0?
```golang
	if t.LpSupply.IsZero() {
		return ErrDenominatorZero
	}
```

PR #392, opened 2024-03-15
curve-stable-meta-ng: simulator

it4rb:

### DIFF
--- a/pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
+++ b/pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
@@ -1,0 +1,251 @@
+package stablemetang
+
+import (
+	"fmt"
+
+	"github.com/KyberNetwork/blockchain-toolkit/number"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
+	stableng "github.com/KyberNetwork/kyberswap-dex-lib/pkg/liquidity-source/curve/stable-ng"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/curve"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
+	"github.com/KyberNetwork/logger"
+	"github.com/holiman/uint256"
+)
+
+// ICurveBasePool is the interface for curve base pool inside a meta pool
+// this is slightly different to the one in SC (return fee)
+type ICurveBasePool interface {
+	GetInfo() pool.PoolInfo
+	GetTokenIndex(address string) int
+
+	GetVirtualPriceU256(vPrice *uint256.Int, D *uint256.Int) error
+
+	CalculateTokenAmountU256(amounts []uint256.Int, deposit bool, mintAmount *uint256.Int, feeAmounts []uint256.Int) error
+	CalculateWithdrawOneCoinU256(tokenAmount *uint256.Int, i int, dy *uint256.Int, dyFee *uint256.Int) error
+
+	// similar to RemoveLiquidityOneCoinU256, but pass in result from CalculateWithdrawOneCoinU256
+	ApplyRemoveLiquidityOneCoinU256(i int, tokenAmount, dy, dyFee *uint256.Int) error
+
+	// similar to AddLiquidity, but pass in result from CalculateTokenAmountU256
+	ApplyAddLiquidity(amounts, feeAmounts []uint256.Int, mintAmount *uint256.Int) error
+}
+
+// the normal swap at meta pool is identical to stable-ng,
+// so we'll inherit from stableng.PoolSimulator to reuse its methods
+type PoolSimulator struct {
+	stableng.PoolSimulator
+	basePool ICurveBasePool
+}
+
+func NewPoolSimulator(entityPool entity.Pool, basePool ICurveBasePool) (*PoolSimulator, error) {
+	sim, err := stableng.NewPoolSimulator(entityPool)
+	if err != nil {
+		return nil, err
+	}
+
+	return &PoolSimulator{*sim, basePool}, err
+}
+
+func (t *PoolSimulator) CalcAmountOut(param pool.CalcAmountOutParams) (*pool.CalcAmountOutResult, error) {
+	tokenAmountIn := param.TokenAmountIn
+	tokenOut := param.TokenOut
+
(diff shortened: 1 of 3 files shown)


pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
```diff
@@ -0,0 +1,251 @@
...
+	// update balance from the meta swap component
+	metaInfo := swapInfo.Meta
+	t.Reserves[metaInfo.TokenInIndex].Add(&t.Reserves[metaInfo.TokenInIndex], &metaInfo.AmountIn)
+	number.FillBig(&t.Reserves[metaInfo.TokenInIndex], t.Info.Reserves[metaInfo.TokenInIndex])
+
+	t.Reserves[metaInfo.TokenOutIndex].Sub(&t.Reserves[metaInfo.TokenOutIndex], number.Add(&metaInfo.AmountOut, &metaInfo.AdminFee))
+	number.FillBig(&t.Reserves[metaInfo.TokenOutIndex], t.Info.Reserves[metaInfo.TokenOutIndex])
+
+	// if output coin is from base pool
+	withDraw := swapInfo.Withdraw
```
hieuvo, line 159: minor typo 
```suggestion
	withdraw := swapInfo.Withdraw
```

pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
```diff
@@ -0,0 +1,251 @@
...
+	if fromId >= 0 && toId >= 0 {
+		return curve.Meta{
+			TokenInIndex:  fromId,
+			TokenOutIndex: toId,
+			Underlying:    false,
+		}
+	}
+	var baseFromId = t.getUnderlyingIndex(tokenIn)
+	var baseToId = t.getUnderlyingIndex(tokenOut)
+	return curve.Meta{
```
hieuvo, line 222: we still return even if 1 of baseFromId and baseToId is negative?

pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
```diff
@@ -0,0 +1,251 @@
...
+		TokenInIndex:  baseFromId,
+		TokenOutIndex: baseToId,
+		Underlying:    true,
+	}
+}
+
+func (t *PoolSimulator) GetTokens() []string {
+	var result []string
```
hieuvo, line 230: minor 
```suggestion
	result := make([]string, 0, len(t.GetInfo().Tokens)+len(t.BasePool.GetInfo().Tokens))
```

pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
it4rb: this function only get called when encoding route, so it's very likely that the CalcAmountOut has been called and succeeded before with the same parameter, so I think here we can assume that they're correct

yes there will be cases where the route info supplied is not correct, but in those cases the encoded data will likely be incorrect anyway

pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
```diff
@@ -0,0 +1,251 @@
...
+		return &pool.CalcAmountOutResult{}, ErrTokenToUnderLyingNotSupported
+	}
+
+	if tokenIndexFrom >= 0 && tokenIndexTo >= 0 {
+		// this is normal swap at meta pool, reuse the method from stable-ng
+		return t.PoolSimulator.CalcAmountOut(param)
+	}
+
+	// swap between meta coins and base pool's coins
+	var baseInputIndex = t.basePool.GetTokenIndex(tokenAmountIn.Token)
```
hieuvo, line 67: can use just call `t.getUnderlyingIndex`?

pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
```diff
@@ -0,0 +1,251 @@
...
+func (t *PoolSimulator) CalcAmountOut(param pool.CalcAmountOutParams) (*pool.CalcAmountOutResult, error) {
+	tokenAmountIn := param.TokenAmountIn
+	tokenOut := param.TokenOut
+
+	var tokenIndexFrom = t.Info.GetTokenIndex(tokenAmountIn.Token)
+	var tokenIndexTo = t.Info.GetTokenIndex(tokenOut)
+
+	// cannot swap between the last meta coin and base pool's coins (because the last coin is LPtoken of base pool)
+	if (tokenIndexFrom == t.NumTokens-1 && tokenIndexTo < 0) || (tokenIndexTo == t.NumTokens-1 && tokenIndexFrom < 0) {
+		return &pool.CalcAmountOutResult{}, ErrTokenToUnderLyingNotSupported
```
hieuvo, line 58: minor, typo
```suggestion
		return &pool.CalcAmountOutResult{}, ErrTokenToUnderlyingNotSupported
```

pkg/liquidity-source/curve/stable-meta-ng/pool_simulator.go
hieuvo, line 67: consider to refactor `getUnderlyingIndex` method a little bit, then we can reuse its logic
```golang
func (t *Pool) getUnderlyingIndex(token string) int {
	var tokenIndex = t.GetTokenIndex(token)
	if tokenIndex >= 0 {
		return tokenIndex
	}
	return t.getBaseUnderlyingIndex(tokenIndex, token)
}

func (t *Pool) getBaseUnderlyingIndex(tokenIndex int, token string) int {
	var baseIndex = t.BasePool.GetTokenIndex(token)
	var maxCoin = len(t.Info.Tokens) - 1
	if tokenIndex < 0 && baseIndex >= 0 {
		tokenIndex = maxCoin + baseIndex
	}
	return tokenIndex
}

```

```suggestion
	var baseInputIndex = t.getBaseUnderlyingIndex(tokenIndexFrom, tokenAmountIn.Token)
```

it4rb: um here we need the raw result of `basePool.GetTokenIndex` to determine if both of the coins are from base pool, so I think we can't use `getBaseUnderlyingIndex` here (maybe we can modify it to return a flag to indicate whether it's a base coin, but then it's a bit overcomplicated)

PR #400, opened 2024-03-22
curve tricrypto-ng: pool tracker

it4rb:

### DIFF
--- a/pkg/liquidity-source/curve/tricrypto-ng/pool_tracker.go
+++ b/pkg/liquidity-source/curve/tricrypto-ng/pool_tracker.go
@@ -54,10 +54,12 @@ func (t *PoolTracker) GetNewPoolState(
 
 		balances = make([]*big.Int, len(p.Tokens))
 
+		numDepCoins = len(p.Tokens) - 1 // other coins will have price based on the 1st coin
+
 		// These 3 slices only has length = number of tokens - 1 (check in the contract)
-		priceScales  = make([]*big.Int, len(p.Tokens)-1)
-		priceOracles = make([]*big.Int, len(p.Tokens)-1)
-		lastPrices   = make([]*big.Int, len(p.Tokens)-1)
+		priceScales  = make([]*big.Int, numDepCoins)
+		priceOracles = make([]*big.Int, numDepCoins)
+		lastPrices   = make([]*big.Int, numDepCoins)
 	)
 
 	calls := t.ethrpcClient.NewRequest().SetContext(ctx)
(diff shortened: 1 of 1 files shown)


hanguyenk (approved):

pkg/liquidity-source/curve/tricrypto-ng/pool_tracker.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		lg.WithFields(logger.Fields{"error": err}).Error("failed to marshal extra data")
+		return entity.Pool{}, err
+	}
+
+	var reserves = make(entity.PoolReserves, 0, len(balances)+1)
+	for i := range balances {
+		reserves = append(reserves, balances[i].String())
+	}
+
```
hieuvo, line 236: don't we append `lpSupply` to `reserves`?

pkg/liquidity-source/curve/tricrypto-ng/pool_tracker.go
```diff
@@ -0,0 +1,242 @@
...
+		MidFee:              number.SetFromBig(midFee),
+		OutFee:              number.SetFromBig(outFee),
+		LpSupply:            number.SetFromBig(lpSupply),
+		XcpProfit:           number.SetFromBig(xcpProfit),
+		VirtualPrice:        number.SetFromBig(virtualPrice),
+		AllowedExtraProfit:  number.SetFromBig(allowedExtraProfit),
+		AdjustmentStep:      number.SetFromBig(adjustmentStep),
+	}
+	extra.PriceScale = make([]uint256.Int, len(priceScales))
+	lo.ForEach(priceScales, func(item *big.Int, i int) { extra.PriceScale[i].SetFromBig(item) })
```
hieuvo, line 220: just wonder why we have to use `lo.ForEach`, while a simple for loop works well here?

pkg/liquidity-source/curve/tricrypto-ng/pool_tracker.go
it4rb: um no, we have it in `extra` already, I don't really know why we put that in to reserves previously

pkg/liquidity-source/curve/tricrypto-ng/pool_tracker.go
it4rb: um no specific reason, I just find them more familiar to me (coming from Scala background) and less code than the verbose loop

I've been bitten by some bugs in handwritten loop before, for example this code to split chunk (https://github.com/KyberNetwork/limit-order/pull/223/files#diff-d790a627480c897fa30fb5220132d0c44c4ef84fcc498155633d10ffe71de059R287). I know the bug is simple and could have been avoided. But I think if we had used `lo` at the beginning then we don't even have to think about that

PR #469, opened 2024-06-10
update balance swaap-v2

tien7668:
update price levels of swaap-v2


hanguyenk (approved):

pkg/liquidity-source/swaap-v2/pool_simulator.go
```diff
@@ -204,3 +213,25 @@ func getAmountOut(amountIn float64, priceLevels []PriceLevel) (float64, error) {
...
 }
+
+func getNewPriceLevelsState(amountIn float64, priceLevels []PriceLevel) []PriceLevel {
+	if len(priceLevels) == 0 {
+		return priceLevels
+	}
+
+	for i, priceLevel := range priceLevels {
+		if amountIn > priceLevel.Level {
+			continue
```
NgoKimPhu, line 224: this `continue` won't go to line 231?

NgoKimPhu: oh looks like swaap-v2 price levels are cumulative

pkg/liquidity-source/swaap-v2/pool_simulator.go
```diff
@@ -204,3 +213,25 @@ func getAmountOut(amountIn float64, priceLevels []PriceLevel) (float64, error) {
...
+		} else if amountIn == priceLevel.Level {
+			if i == len(priceLevels)-1 {
+				return nil
+			}
+			return priceLevels[i+1:]
```
NgoKimPhu, line 229: in other RFQ's I see they often use first price level as the minimum swap amount, but looks like swaap doesn't have this?

pkg/liquidity-source/swaap-v2/pool_simulator.go
tien7668: No, It won't. swaap-v2's price-levels are accumulated, mean at index i, the level[i] = level[i-1] + amount[i]

pkg/liquidity-source/swaap-v2/pool_simulator.go
tien7668: Yeah, haven't found any info about minimum amount from the docs and current impl. Actually, first price level we got is 0, means we can swap with any amount, I believe

PR #486, opened 2024-07-10
EX-1854 Add Ambient for AEVM

phqb:


pkg/liquidity-source/ambient/types.go
```diff
@@ -0,0 +1,34 @@
...
+	LastCreateTime uint64 `json:"lastCreateTime"`
+}
+
+type StaticExtra struct {
+	// Base token, 0x0 if native token
+	Base string `json:"base"`
+	// Quote token
+	Quote string `json:"quote"`
+	// The index or discriminator of the pool in a logical list of pools with (base, quote) tokens. Default is "420".
+	PoolIdx string `json:"pool_idx"`
```
piavgh, line 26: we should use camelCase here, same with SwapAddress below

pkg/liquidity-source/ambient/types.go
```diff
@@ -0,0 +1,34 @@
+package ambient
+
+type FetchPoolsResponse struct {
+	Pools []Pool `json:"pools"`
+}
+
+type Pool struct {
+	ID          string `json:"id"`
+	BlockCreate string `json:"blockCreate"`
+	TimeCreate  uint64 `json:"timeCreate,string"`
```
piavgh, line 10: what is the purpose of the `,string` here?

pkg/liquidity-source/ambient/types.go
phqb: done

pkg/liquidity-source/ambient/types.go
phqb: It tells json to unmarshal string as number, for example `{"timeCreate":"123"}` will be unmarshaled as `Pool{ TimeCreate: 123 }`.

piavgh (approved):

PR #493, opened 2024-07-24
integrate traderjoe v2.2

sunspirit9999:


pkg/source/liquiditybookv20/pool_tracker.go
```diff
@@ -234,6 +234,15 @@ func (d *PoolTracker) querySubgraph(ctx context.Context, p entity.Pool) (*queryS
...
 		}
+
+		// if no bin returned, stop
+		if resp.Pair == nil || len(resp.Pair.Bins) == 0 {
+			logger.WithFields(logger.Fields{
+				"poolAddress": p.Address,
+			}).Info("no bin returned")
+			break
+		}
+
```
lehainam-dev, line 245: Can you explain more about this change?

sunspirit9999: we should check resp.Pair != nil before use its properties, btw we can avoid to be panic (it really happened when i tested on my local)

lehainam-dev: Oh I see. Yeah I agree.

sunspirit9999: maybe due to the subgraph is still syncing

lehainam-dev (approved):

PR #498, opened 2024-08-21
fix: dodo v2 pool panic when cloning

piavgh:


it4rb (approved):
lgtm, please consider removing mutex in the sim struct as well

it4rb (approved):

PR #508, opened 2024-09-16
feat: integrate lite-psm

piavgh:

### DIFF
--- a/pkg/liquidity-source/litepsm/constant.go
+++ b/pkg/liquidity-source/litepsm/constant.go
@@ -25,8 +25,5 @@ var (
 	WAD        = number.Number_1e18
 
 	// 2^256 - 1 (type(uint256).max)
-	HALTED = new(uint256.Int).Sub(
-		new(uint256.Int).Lsh(number.Number_1, 256),
-		number.Number_1,
-	)
+	HALTED = new(uint256.Int).SetAllOne()
 )
--- a/pkg/liquidity-source/litepsm/constant.go
+++ b/pkg/liquidity-source/litepsm/constant.go
@@ -1,0 +1,32 @@
+package litepsm
+
+import (
+	"github.com/KyberNetwork/blockchain-toolkit/number"
+	"github.com/holiman/uint256"
+)
+
+const DexTypeLitePSM = "lite-psm"
+
+const (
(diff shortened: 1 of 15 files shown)


pkg/liquidity-source/litepsm/constant.go
```diff
@@ -0,0 +1,32 @@
...
+
+var (
+	DefaultGas = Gas{SellGem: 115000, BuyGem: 115000}
+	WAD        = number.Number_1e18
+
+	// 2^256 - 1 (type(uint256).max)
+	HALTED = new(uint256.Int).Sub(
+		new(uint256.Int).Lsh(number.Number_1, 256),
+		number.Number_1,
+	)
```
NgoKimPhu, line 31: minor
```suggestion
	HALTED = new(uint256.Int).SetAllOne()
```

pkg/liquidity-source/litepsm/pool_simulator.go
```diff
@@ -0,0 +1,135 @@
...
+			},
+			Fee: &pool.TokenAmount{
+				Token:  tokenOut,
+				Amount: fee.ToBig(),
+			},
+			Gas: p.gas.BuyGem,
+		}, nil
+	}
+
+	gemAmt, fee, err := p.litePSM.sellGem(amountIn)
```
NgoKimPhu, line 98: do we need to verify that token in is gem?

pkg/liquidity-source/litepsm/pool_tracker.go
```diff
@@ -0,0 +1,156 @@
+package litepsm
+
+import (
+	"context"
+	"encoding/json"
```
NgoKimPhu, line 5: (side discussion) I rmb we used to use goccy, is it not worth it?

NgoKimPhu (approved):

pkg/liquidity-source/litepsm/pool_simulator.go
piavgh: valid point. I added a check

pkg/liquidity-source/litepsm/constant.go
piavgh: thanks, I didn't know that method SetAllOne exists :D

PR #509, opened 2024-09-17
feat: integrate curve twocrypto-ng

NgoKimPhu:


pkg/liquidity-source/curve/tricrypto-ng/curve_tricrypto_optimized_weth.go
```diff
@@ -73,14 +77,14 @@ func (t *PoolSimulator) GetDy(
...
-		yOrg.Div(number.SafeMul(yOrg, &t.Extra.PriceScale[j-1]), U_1e18)
+		yOrg.Div(number.SafeMul(yOrg, &t.Extra.PriceScale[j-1]), Precision)
 	}
 	xp[j].Set(yOrg)
 
 	return nil
 }
 
-// GetDx https://github.com/curvefi/tricrypto-ng/blob/c4093cbda18ec8f3da21bf7e40a3f8d01c5c4bd3/contracts/main/CurveCryptoViews3Optimized.vy#L76
+// GetDx https://github.com/curvefi/twocrypto-ng/blob/1c800bd/contracts/main/CurveCryptoViews2Optimized.vy#L79
```
piavgh, line 87: this file is the implmentation of tricrypto. Is it ok to change the source liquidity code to twocrypto?

pkg/liquidity-source/curve/twocrypto-ng/pool_simulator.go
```diff
@@ -0,0 +1,213 @@
...
+			Reserves:   reservesBI,
+			Checked:    false,
+		},
+	}
+
+	sim.gas = DefaultGas
+
+	return sim, nil
+}
+var zz = map[string]bool{}
```
piavgh, line 89: why do we have this?

pkg/liquidity-source/curve/twocrypto-ng/pool_simulator.go
```diff
@@ -0,0 +1,213 @@
...
+			},
+			Fee: &pool.TokenAmount{
+				Token:  tokenOut,
+				Amount: fee.ToBig(),
+			},
+			Gas:      t.gas,
+			SwapInfo: swapInfo,
+		}, nil
+	}
+	return &pool.CalcAmountOutResult{}, fmt.Errorf("tokenIndexFrom %v or tokenIndexTo %v is not correct", tokenIndexFrom, tokenIndexTo)
```
piavgh, line 128: I think we can rewrite the code like this

```suggestion
	if tokenIndexFrom < 0 || tokenIndexTo < 0 {
		return &pool.CalcAmountOutResult{}, fmt.Errorf("tokenIndexFrom %v or tokenIndexTo %v is not correct", tokenIndexFrom, tokenIndexTo)
	}

	var amountOut, fee, amount uint256.Int
	amount.SetFromBig(tokenAmountIn.Amount)

	swapInfo := SwapInfo{}
	err := t.GetDy(
		tokenIndexFrom,
		tokenIndexTo,
		&amount,
		&amountOut, &fee, &swapInfo.K0, swapInfo.Xp[:],
	)
	if err != nil {
		return &pool.CalcAmountOutResult{}, err
	}

	if amountOut.IsZero() {
		return &pool.CalcAmountOutResult{}, ErrZero
	}

	return &pool.CalcAmountOutResult{
		TokenAmountOut: &pool.TokenAmount{
			Token:  tokenOut,
			Amount: amountOut.ToBig(),
		},
		Fee: &pool.TokenAmount{
			Token:  tokenOut,
			Amount: fee.ToBig(),
		},
		Gas:      t.gas,
		SwapInfo: swapInfo,
	}, nil
```

pkg/liquidity-source/curve/twocrypto-ng/pool_simulator.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (t *PoolSimulator) CalcAmountIn(param pool.CalcAmountInParams) (*pool.CalcAmountInResult, error) {
+	tokenIn := param.TokenIn
+	tokenAmountOut := param.TokenAmountOut
+
+	var tokenIndexFrom = t.Info.GetTokenIndex(tokenIn)
+	var tokenIndexTo = t.Info.GetTokenIndex(tokenAmountOut.Token)
+
+	if tokenIndexFrom >= 0 && tokenIndexTo >= 0 {
```
piavgh, line 138: same here

pkg/liquidity-source/curve/twocrypto-ng/pool_tracker.go
```diff
@@ -0,0 +1,245 @@
...
+			ABI:    curveTwocryptoNGABI,
+			Target: p.Address,
+			Method: poolMethodLastPrices,
+			Params: nil,
+		}, []interface{}{&lastPrices[i]})
+	}
+	if res, err := calls.TryBlockAndAggregate(); err != nil {
+		lg.WithFields(logger.Fields{"error": err}).Error("failed to aggregate call pool data")
+		return entity.Pool{}, err
+	} else if res.BlockNumber != nil {
```
piavgh, line 200: Because in case of error, the code returns immediatelly, I think we can separate the second if block

```suggestion
	}
    
    if res.BlockNumber != nil {
```

piavgh:
also, please fix linting issue
For code that we don't use but we still want to keep it as reference, we can add this comment
```
// nolint: unused
```

pkg/liquidity-source/curve/tricrypto-ng/curve_tricrypto_optimized_weth.go
NgoKimPhu: oops

pkg/liquidity-source/curve/twocrypto-ng/pool_simulator.go
NgoKimPhu: haha i forgot to remove, was using this to increase test coverage

pkg/liquidity-source/curve/twocrypto-ng/pool_simulator.go
NgoKimPhu: copied from tricrypto, let me update for both

pkg/liquidity-source/curve/twocrypto-ng/pool_tracker.go
NgoKimPhu: separating the second if block in this case doesn't reduce the indent level of the nested block and makes the code longer though? moreover in this case, `res` is scoped within the `if`

NgoKimPhu:
check failed tests

piavgh (approved):

PR #529, opened 2024-10-02
fix : Add check available reserve and adjust default gas

sunspirit99:

### DIFF
--- a/pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
+++ b/pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
@@ -26,7 +26,7 @@ type PoolSimulator struct {
 }
 
 var (
-	defaultGas = Gas{Liquidate: 1000000}
+	defaultGas = Gas{Liquidate: 1250000}
 )
 
 func NewPoolSimulator(entityPool entity.Pool) (*PoolSimulator, error) {
--- a/pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
+++ b/pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
@@ -73,7 +73,7 @@ func (s *PoolSimulator) CalcAmountOut(param poolpkg.CalcAmountOutParams) (*poolp
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
```diff
@@ -116,3 +127,13 @@ func (s *PoolSimulator) CanSwapTo(address string) []string {
 
 	return result
 }
+
+// Helper function to get reserve for a specific token
+func (s *PoolSimulator) getReserveForToken(token string) (*big.Int, error) {
+	for i, t := range s.GetTokens() {
```
NgoKimPhu, line 133: Use GetTokenIndex

pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
```diff
@@ -66,6 +68,15 @@ func (s *PoolSimulator) CalcAmountOut(param poolpkg.CalcAmountOutParams) (*poolp
 
 	tokenAmountOut = new(big.Int).Div(tokenAmountOut, divisor1e27)
 
+	reserveTokenOut, err := s.getReserveForToken(param.TokenOut)
+	if err != nil {
+		return nil, err
+	}
+
+	if tokenAmountOut.Cmp(reserveTokenOut) >= 0 {
```
NgoKimPhu, line 76: Is it not okay to be equal?

pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
sunspirit99: 8cfe6ed351c9baf235c5c3b46ea44b108d638ed6

pkg/liquidity-source/fluid/vault-t1/pool_simulator.go
sunspirit99: yes, it can be :)) f635c16378ee4e3f71b55d6566ef96f3b82f7e6c

NgoKimPhu (approved):

PR #543, opened 2024-10-14
feat: ✨ bebop: add inventory

thanhpp:
## Why did we need it?

- Limit the token's amount through bebop

## Related Issue

- https://team-kyber.slack.com/archives/C061UNZDUVC/p1728637720886349?thread_ts=1727247781.998879&cid=C061UNZDUVC


pkg/liquidity-source/bebop/inventory.go
```diff
@@ -0,0 +1,67 @@
+package bebop
+
+import (
+	"math/big"
+	"sync"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
+)
+
+type Inventory struct {
```
NgoKimPhu, line 10: seems there's no change vs `kyberpmm.NewInventory`, can we just reuse that?

pkg/liquidity-source/bebop/limit.go
```diff
@@ -0,0 +1,30 @@
+package bebop
+
+import "math/big"
+
+// Limit: limit 1 bebop swap per route
+type Limit struct {
+	hasSwap bool
```
NgoKimPhu, line 7: use `atomic.Bool`

thanhpp: updated in https://github.com/KyberNetwork/kyberswap-dex-lib/pull/543/commits/4e0d37ce6e4c83261d10ecb157169bd54f1ce299

NgoKimPhu (approved):

PR #546, opened 2024-10-16
EX-2270: integrate clipper rfq

lehainam-dev:

### DIFF
--- a/pkg/liquidity-source/clipper/client/http.go
+++ b/pkg/liquidity-source/clipper/client/http.go
@@ -1,0 +1,73 @@
+package client
+
+import (
+	"context"
+	"errors"
+
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/liquidity-source/clipper"
+	"github.com/go-resty/resty/v2"
+)
+
+const (
+	quotePath = "/rfq/quote"
+	signPath  = "/rfq/sign"
+)
+
+var (
+	ErrQuoteFailed = errors.New("quote failed")
+	ErrSignFailed  = errors.New("sign failed")
+)
+
+type httpClient struct {
+	client *resty.Client
+	config clipper.HTTPClientConfig
+}
+
+func NewHTTPClient(config clipper.HTTPClientConfig) *httpClient {
+	client := resty.New().
+		SetBaseURL(config.BaseURL).
+		SetTimeout(config.Timeout.Duration).
+		SetRetryCount(config.RetryCount).
+		SetHeader("Authorization", "Basic "+config.BasicAuthKey)
+
+	return &httpClient{
+		client: client,
+		config: config,
+	}
+}
+
+func (c *httpClient) RFQ(ctx context.Context, params clipper.QuoteParams) (clipper.SignResponse, error) {
+	// 1. Call quote endpoint
+	req := c.client.R().SetContext(ctx).SetBody(params)
+
+	var quoteRes clipper.QuoteResponse
+	res, err := req.SetResult(&quoteRes).Post(quotePath)
+	if err != nil {
+		return clipper.SignResponse{}, err
+	}
+
+	if !res.IsSuccess() {
+		return clipper.SignResponse{}, ErrQuoteFailed
+	}
+
+	// 2. Call sign endpoint with `quote_id` received from step 1
+	req = c.client.R().SetContext(ctx).SetBody(clipper.SignParams{
+		QuoteID:            quoteRes.ID,
+		DestinationAddress: params.DestinationAddress,
+		SenderAddress:      params.SenderAddress,
+		NativeInput:        false,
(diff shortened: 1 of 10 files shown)


pkg/liquidity-source/clipper/client/http.go
```diff
@@ -0,0 +1,73 @@
...
+		config: config,
+	}
+}
+
+func (c *httpClient) RFQ(ctx context.Context, params clipper.QuoteParams) (clipper.SignResponse, error) {
+	// 1. Call quote endpoint
+	req := c.client.R().SetContext(ctx).SetBody(params)
+
+	var quoteRes clipper.QuoteResponse
+	res, err := req.SetResult(&quoteRes).Post(quotePath)
```
NgoKimPhu, line 44: if the response has status > 299, need to also `SetError(&quoteRes)` to read the body

pkg/liquidity-source/clipper/client/http.go
```diff
@@ -0,0 +1,73 @@
...
+	req = c.client.R().SetContext(ctx).SetBody(clipper.SignParams{
+		QuoteID:            quoteRes.ID,
+		DestinationAddress: params.DestinationAddress,
+		SenderAddress:      params.SenderAddress,
+		NativeInput:        false,
+		NativeOutput:       false,
+	})
+
+	var signRes clipper.SignResponse
+	res, err = req.SetResult(&signRes).Post(signPath)
```
NgoKimPhu, line 63: same here

pkg/liquidity-source/clipper/client/http.go
```diff
@@ -0,0 +1,73 @@
...
+	req := c.client.R().SetContext(ctx).SetBody(params)
+
+	var quoteRes clipper.QuoteResponse
+	res, err := req.SetResult(&quoteRes).Post(quotePath)
+	if err != nil {
+		return clipper.SignResponse{}, err
+	}
+
+	if !res.IsSuccess() {
+		return clipper.SignResponse{}, ErrQuoteFailed
```
NgoKimPhu, line 50: parsing and mapping the error is optional but we should still have some logs to know why it fails

pkg/liquidity-source/clipper/pool_simulator.go
```diff
@@ -0,0 +1,167 @@
...
+			new(big.Float).Mul(M, inX),
+			pX,
+		),
+		pY,
+	)
+	outY.Mul(outY, bignumber.TenPowDecimals(assetOut.Decimals))
+
+	amountOut, _ := outY.Int(nil)
+
+	// TODO: Apply the final check FMV
```
NgoKimPhu, line 141: TODO?

pkg/liquidity-source/clipper/pool_simulator.go
```diff
@@ -0,0 +1,167 @@
...
+	M := new(big.Float).SetFloat64((basisPoint - pairInfo.FeeInBasisPoints) / basisPoint)
+
+	outY := new(big.Float).Quo(
+		new(big.Float).Mul(
+			new(big.Float).Mul(M, inX),
+			pX,
+		),
+		pY,
+	)
+	outY.Mul(outY, bignumber.TenPowDecimals(assetOut.Decimals))
```
NgoKimPhu, line 137: (just a comment) so many `new(big.Float)` 😭

pkg/liquidity-source/clipper/pool_simulator.go
```diff
@@ -0,0 +1,167 @@
...
+		TokenAmountOut: &pool.TokenAmount{
+			Token:  assetOut.Address,
+			Amount: amountOut,
+		},
+		Fee: &pool.TokenAmount{
+			Token:  assetOut.Address,
+			Amount: bignumber.ZeroBI,
+		},
+
+		// TODO: Define gas
```
NgoKimPhu, line 153: gas gas gas

pkg/liquidity-source/clipper/client/http.go
lehainam-dev: Updated

pkg/liquidity-source/clipper/pool_simulator.go
lehainam-dev: The FMV will always return true, I added a comment to explain.

pkg/liquidity-source/clipper/pool_simulator.go
lehainam-dev: I move to the local variable `big.Float`, so the data can be allocated in stack instead.

pkg/liquidity-source/clipper/pool_simulator.go
lehainam-dev: Updated

NgoKimPhu (approved):

PR #551, opened 2024-10-22
feat: integrate frxeth, meth, and ondo-usdy

minhnhathoang:


pkg/liquidity-source/mantle/meth/constant.go
```diff
@@ -0,0 +1,30 @@
...
+	mantleLSPStakingMethodMaximumDepositAmount   = "maximumDepositAmount"
+	mantleLSPStakingMethodMinimumStakeBound      = "minimumStakeBound"
+	mantleLSPStakingMethodMaximumMETHSupply      = "maximumMETHSupply"
+
+	mETHMethodTotalSupply = "totalSupply"
+)
+
+var (
+	defaultGas = Gas{
+		Stake: 100000,
```
NgoKimPhu, line 28:
```suggestion
		Stake: 130000,
```

pkg/liquidity-source/mantle/meth/math.go
```diff
@@ -0,0 +1,67 @@
...
+
+var (
+	// 2^256 - 1
+	maxUint256 = new(uint256.Int).Sub(
+		new(uint256.Int).Lsh(number.Number_1, 256),
+		number.Number_1,
+	)
+)
+
+func mulDiv(x, y, denominator *uint256.Int) (*uint256.Int, error) {
```
NgoKimPhu, line 16: replace this whole thing with `.MulDivOverflow`

pkg/liquidity-source/mantle/meth/math.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/KyberNetwork/blockchain-toolkit/number"
+	"github.com/holiman/uint256"
+)
+
+var (
+	// 2^256 - 1
+	maxUint256 = new(uint256.Int).Sub(
+		new(uint256.Int).Lsh(number.Number_1, 256),
+		number.Number_1,
+	)
```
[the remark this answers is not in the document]
minhnhathoang: simplify max uint256 calculation

pkg/liquidity-source/mantle/meth/pool_simulator.go
```diff
@@ -0,0 +1,148 @@
...
+	amountOut, err := s.ethToMETH(amountIn)
+	if err != nil {
+		return nil, err
+	}
+
+	if new(uint256.Int).Add(amountOut, s.mETHTotalSupply).Cmp(s.maximumMETHSupply) > 0 {
+		return nil, ErrMaximumMETHSupplyExceeded
+	}
+
+	if amountOut.Cmp(uint256.NewInt(0)) < 0 {
```
NgoKimPhu, line 98: this should compare with `minMETHAmount`, not `0`

minhnhathoang: `minMETHAmount` is a fixed argument

https://github.com/KyberNetwork/ks-dex-aggregator-sc/blob/develop/src/contracts/executor-helpers/ExecutorHelper6.sol#L141

pkg/liquidity-source/mantle/meth/pool_simulator.go
```diff
@@ -0,0 +1,148 @@
...
+
+	mETHSupplyAdjusted := new(uint256.Int).Mul(s.mETHTotalSupply, uint256.NewInt(uint64(common.UInt16BasisPoints-s.exchangeAdjustmentRate)))
+	totalControlledAdjusted := new(uint256.Int).Mul(s.totalControlled, uint256.NewInt(uint64(common.UInt16BasisPoints)))
+
+	amountOut, err := mulDiv(mETHAmount, mETHSupplyAdjusted, totalControlledAdjusted)
+	if err != nil {
+		return nil, err
+	}
+
+	return amountOut, nil
```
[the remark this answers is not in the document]
minhnhathoang: Appreciate it! It is very useful to know.

In this case, the pointer mETHAmount is passed from CalcAmountOut as `amountIn := uint256.MustFromBig(params.TokenAmountIn.Amount)` and after `ethToMETH` is called, we no longer need to use `amountIn`, so we can write `amountOut, overflow := mETHAmount.MulDivOverflow(mETHAmount, &mETHSupplyAdjusted, &totalControlledAdjusted)`. However, this could lead to an ambiguous effect of changing amountIn, even though AmountIn is no longer used after calling the ethToMETH function. In this case, should we directly use the pointer `mETHAmount`?

pkg/liquidity-source/mantle/meth/pool_simulator.go
NgoKimPhu: ah I see. anyway can update this to 

```suggestion
	if amountOut.Sign() < 0 {
```

to save on 1 allocation

pkg/liquidity-source/mantle/meth/pool_simulator.go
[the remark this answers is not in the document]
NgoKimPhu: in this case we should not, because it's best not to make any assumptions about whether the arguments are reused or not. For example, in strict defensive codes, input arrays are usually even copied before use.

pkg/liquidity-source/ondo-usdy/constant.go
```diff
@@ -0,0 +1,19 @@
...
+
+const (
+	mUSDMethodPaused                   = "paused"
+	rwaDynamicOracleMethodGetPriceData = "getPriceData"
+)
+
+var (
+	defaultGas = Gas{
+		Wrap:   550000000,
+		Unwrap: 550000000,
```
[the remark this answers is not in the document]
NgoKimPhu: currently we provide gas based on eth, for mantle looks like we haven't handled special gas logic here, so let's leave it the usual amount on eth

pkg/liquidity-source/mantle/meth/constant.go
```diff
@@ -0,0 +1,30 @@
...
+	mantleLSPStakingMethodMaximumDepositAmount   = "maximumDepositAmount"
+	mantleLSPStakingMethodMinimumStakeBound      = "minimumStakeBound"
+	mantleLSPStakingMethodMaximumMETHSupply      = "maximumMETHSupply"
+
+	mETHMethodTotalSupply = "totalSupply"
+)
+
+var (
+	defaultGas = Gas{
+		Stake: 130000,
```
NgoKimPhu, line 28: last time I asked to increase this to 130k, I forgot that there's an overhead for any tx, which is about 30k. so let's reduce this to 100k, thanks and sorry for the trouble

minhnhathoang: should we calculate this number as the average of gasUsed based on the tx receipts? In this case, the average gas is approximately 150,000 (214 transactions from block 21036737 to 20936737 (100,000 blocks)), so should we adjust it to 120,000 instead?

NgoKimPhu (approved):

PR #556, opened 2024-10-28
ft/custom clone

NgoKimPhu:
- **feat: custom clone funcs for pool. implement for uniswap-v3**
- **feat: custom clone funcs for swap limits. refactor to common types**
- **feat: apply uniswapv3 uint256 for pancake-v3**

### DIFF
--- a/pkg/source/pancakev3/pool_simulator.go
+++ b/pkg/source/pancakev3/pool_simulator.go
@@ -8,11 +8,12 @@ import (
 	"strings"
 
 	"github.com/KyberNetwork/logger"
-	"github.com/KyberNetwork/pancake-v3-sdk/constants"
-	v3Entities "github.com/KyberNetwork/pancake-v3-sdk/entities"
-	v3Utils "github.com/KyberNetwork/pancake-v3-sdk/utils"
+	"github.com/KyberNetwork/uniswapv3-sdk-uint256/constants"
+	v3Entities "github.com/KyberNetwork/uniswapv3-sdk-uint256/entities"
+	v3Utils "github.com/KyberNetwork/uniswapv3-sdk-uint256/utils"
 	coreEntities "github.com/daoleno/uniswap-sdk-core/entities"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/holiman/uint256"
 
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
@@ -20,8 +21,7 @@ import (
 )
 
 var (
-	ErrTickNil        = errors.New("tick is nil")
-	ErrV3TicksEmpty   = errors.New("v3Ticks empty")
+	ErrOverflow       = errors.New("bigInt overflow int/uint256")
 	ErrInvalidFeeTier = errors.New("invalid feeTier")
 )
 
@@ -34,7 +34,7 @@ type PoolSimulator struct {
 }
 
 func NewPoolSimulator(entityPool entity.Pool, chainID valueobject.ChainID) (*PoolSimulator, error) {
-	var extra Extra
+	var extra ExtraTickU256
 	if err := json.Unmarshal([]byte(entityPool.Extra), &extra); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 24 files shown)


pkg/source/pool/iface.go
```diff
@@ -41,8 +41,12 @@ type IPoolTracker interface {
 type IPoolSimulator interface {
 	// CalcAmountOut amountOut, fee, gas
 	// the required params is TokenAmountIn and TokenOut.
-	// SwapLimit is optional, individual dex logic will chose to ignore it if it is nill
+	// SwapLimit is optional, individual dex logic will choose to ignore it if it is nil
 	CalcAmountOut(params CalcAmountOutParams) (*CalcAmountOutResult, error)
+	// CloneBalance clones IPoolSimulator to back up old balance state before UpdateBalance by a swap.
+	// Only clones fields updated by UpdateBalance. Returns nil if unimplemented.
+	CloneBalance() IPoolSimulator
```
lehainam-dev, line 48: This naming is hard for me to understand. At first, I though this function clones only Pool's balance (reserve). After reading kyberswap-dex-lib PR, I realize that this function clones the pool state.

How about renaming it to `ClonePoolState()` for better understanding? (`balance` is just a part of pool's state).

pkg/source/pool/pool.go
```diff
@@ -18,6 +18,10 @@ type Pool struct {
 	Info PoolInfo
 }
 
+func (t *Pool) CloneBalance() IPoolSimulator {
+	return nil
+}
```
lehainam-dev, line 23: Should we add default implementation here?
```
  return clone.Slowly(t)
```

pkg/source/pool/iface.go
NgoKimPhu: I call it `CloneBalance()` because of the `UpdateBalance()`, which also actually updates state.

NgoKimPhu: but yeah, let's change to UpdateState()

pkg/source/pool/pool.go
NgoKimPhu: it would be a wrong implementation, because `t` here is only the embedded `Pool`

NgoKimPhu: this can be overcome by using the trick used by redis package, which is to pass the reference of the SpecificPool into Pool during construction, but that requires lots of change

pkg/source/uniswapv3/pool_simulator_test.go
```diff
@@ -262,3 +262,54 @@ func RandNumberString(maxLen int) string {
...
+	var poolSim pool.IPoolSimulator
+	poolSim, err = NewPoolSimulator(*poolEntity, valueobject.ChainIDEthereum)
+	require.NoError(t, err)
+
+	// BenchmarkA-16    	    1153	    978127 ns/op
+	// cloned = clone.Slowly(sim).(pool.IPoolSimulator)
+	// stack overflowed
+	// cloned = clone.Clone(sim).(pool.IPoolSimulator)
+	// BenchmarkA-16    	 5774500	       212.4 ns/op
+	cloned := poolSim.CloneBalance()
```
lehainam-dev, line 280: Nice.

pkg/source/pancakev3/pool_simulator.go
```diff
@@ -133,145 +132,161 @@ func NewPoolSimulator(entityPool entity.Pool, chainID valueobject.ChainID) (*Poo
...
+			return nil, fmt.Errorf("can not GetInputAmount, err: %+v", err)
 		}
-		var totalGas = p.gas.BaseGas + p.gas.CrossInitTickGas*int64(amountOutResult.CrossInitTickLoops)
 
-		if amountOut.Quotient().Cmp(zeroBI) > 0 {
-			return &pool.CalcAmountOutResult{
-				TokenAmountOut: &pool.TokenAmount{
-					Token:  tokenOut,
-					Amount: amountOut.Quotient(),
+		totalGas := p.gas.BaseGas // TODO: update GetInputAmount to return crossed tick if we ever need this
```
lehainam-dev, line 179: Uhm `var totalGas = p.gas.BaseGas + p.gas.CrossInitTickGas*int64(amountOutResult.CrossInitTickLoops)`?
Is there any reason to change this?

lehainam-dev: Oh this logic is from the *big.Int implementation. We will use the *uint256.Int, so no worry if we don't update this.

lehainam-dev: Ok so I gues the uint256 SDK doesn't return `CrossInitTickLoops`.

lehainam-dev (approved):

PR #571, opened 2024-11-05
support algebra ticklens

it4rb:
Algebra doesn't have Ticklens contract deployed, so we need to use a custom one. Currently Algebra doesn't compact the ticktable, so we can only get a small number of ticks for now

This PR also update univ3-clones to support only re-fetching changed ticks (parsed from log)

### DIFF
--- a/pkg/util/ticklens/ticklens.go
+++ b/pkg/util/ticklens/ticklens.go
@@ -8,7 +8,6 @@ import (
 	"sort"
 	"strconv"
 
-	"github.com/daoleno/uniswapv3-sdk/constants"
 	"github.com/daoleno/uniswapv3-sdk/utils"
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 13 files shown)


pkg/util/ticklens/ticklens.go
```diff
@@ -115,6 +142,41 @@ func GetPoolTicksFromSC(
 		}
 	}
 
+	// if we only fetched some ticks, then update them to the original ticks and return
+	if len(changedTicks) > 0 {
+		// ticklens contract might return unchanged tick (in the same word), so need to filter them out
+		changedTickSet := mapset.NewSet(changedTicks...)
```
NgoKimPhu, line 148:
```suggestion
		changedTickSet := mapset.NewThreadUnsafeSet(changedTicks...)
```

pkg/util/ticklens/ticklens.go
```diff
@@ -129,3 +191,17 @@ func GetPoolTicksFromSC(
 func getTickSpacing(swapFee float64) int {
 	return constants.TickSpacings[constants.FeeAmount(swapFee)]
```
NgoKimPhu, line 192: this one will be wrong for pancake-v3 which have a different tick spacing for fee=2500

NgoKimPhu (approved):

PR #587, opened 2024-11-13
fix(KTX): Add configuration for GMX dex-type to select correct contract function on Mantle

sunspirit99:

### DIFF
--- a/pkg/source/gmx/price_feed_reader.go
+++ b/pkg/source/gmx/price_feed_reader.go
@@ -5,18 +5,21 @@ import (
 	"math/big"
 
 	"github.com/KyberNetwork/ethrpc"
+	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/valueobject"
 	"github.com/KyberNetwork/logger"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 )
 
 type PriceFeedReader struct {
+	chainId      valueobject.ChainID
 	abi          abi.ABI
 	ethrpcClient *ethrpc.Client
 	log          logger.Logger
 }
 
-func NewPriceFeedReader(ethrpcClient *ethrpc.Client) *PriceFeedReader {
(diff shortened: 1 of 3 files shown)


pkg/source/gmx/price_feed_reader.go
```diff
@@ -47,10 +52,15 @@ func (r *PriceFeedReader) getLatestRoundData(ctx context.Context, address string
 
 	rpcRequest := r.ethrpcClient.NewRequest().SetContext(ctx)
 
+	method := priceFeedMethodLatestRoundData
+	if r.chainId == valueobject.ChainIDMantle && r.dexId == string(valueobject.ExchangeKTX) {
+		method = "latestRound"
```
NgoKimPhu, line 57: Is this because of the chain mantle using a different method, or is it because it's a different version?

NgoKimPhu: If it's a different version, this might break other gmx dexes in mantle

pkg/source/gmx/price_feed_reader.go
```diff
@@ -5,18 +5,23 @@ import (
...
 type PriceFeedReader struct {
+	chainId      valueobject.ChainID
+	dexId        string
 	abi          abi.ABI
 	ethrpcClient *ethrpc.Client
 	log          logger.Logger
 }
 
-func NewPriceFeedReader(ethrpcClient *ethrpc.Client) *PriceFeedReader {
+func NewPriceFeedReader(chainID valueobject.ChainID, dexID string, ethrpcClient *ethrpc.Client) *PriceFeedReader {
```
NgoKimPhu, line 21: create a new function to avoid breaking changes. pass in a `priceFeedMethodLatestRoundData` instead of `chainID` and `dexID`

sunspirit99: https://github.com/KyberNetwork/kyberswap-dex-lib/pull/587/commits/47eec775292193b0d4c536c1c01c7bff8df8bcf3

NgoKimPhu (approved):

PR #616, opened 2024-11-26
feat: integrate 1inch limit order v4

piavgh:
## Related Issue

https://team-kyber.atlassian.net/browse/EX-2407

### DIFF
--- a/pkg/liquidity-source/lo1inch/constant.go
+++ b/pkg/liquidity-source/lo1inch/constant.go
@@ -1,0 +1,24 @@
+package lo1inch
+
+import "math/big"
+
+const (
+	DexType = "lo1inch"
+
+	PoolIDPrefix    = "lo1inch"
+	PoolIDSeparator = "_"
+)
+
+var (
+	// BasGas is base gas to executor a tx for LO.
+	BaseGas = 90000
+
+	// GasPerOrderExecutor is gas for executing an order.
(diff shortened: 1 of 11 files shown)


pkg/liquidity-source/lo1inch/constant.go
```diff
@@ -0,0 +1,30 @@
...
+	DexType = "lo1inch"
+
+	PoolIDPrefix    = "lo1inch"
+	PoolIDSeparator = "_"
+
+	// Currently, the total of TVL/reserveUsd in the limit order pool will be very small compared with other pools. So it will be filtered in choosing pools process
+	// We will use big hardcode number to push it into eligible pools for findRoute algorithm.
+	// TODO: when we has correct formula that pool's reserve can be eligible pools.
+	limitOrderPoolReserve    = "10000000000000000000"
+	LimitOrderPoolReserveUSD = 1000000000
```
NgoKimPhu, line 15: reduce this so it's smaller than our LO

NgoKimPhu: actually can we just use the actual reserve?

piavgh: good point, I updated pool-service to use the real reserve

NgoKimPhu (approved):

PR #620, opened 2024-11-28
ft/algebra v1 uint256

NgoKimPhu:
- **feat: use uint256 for algebra-v1**
- **feat: implement CalcAmountIn for algebrav1**


pkg/source/algebrav1/pool_simulator.go
```diff
@@ -18,163 +19,181 @@ import (
...
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("_calculateSwapAndLock failed: %v", err)
+	} else if swapResult.amountCalculated.Sign() <= 0 {
+		return nil, ErrZeroAmountIn
 	}
 
-	return sqrtPriceX96Limit
+	amountIn := swapResult.amountCalculated.ToBig()
+	return &pool.CalcAmountInResult{
```
piavgh, line 97: I think we need to return RemainingTokenAmountOut here, like the RemainingTokenAmountIn in the CalcAmountOut because Algebra-v1 supports partially fill

piavgh: also, Gas is missing

pkg/source/algebrav1/math.go
```diff
@@ -200,5 +196,10 @@ func (p *PoolSimulator) _calculateSwapAndLock(
 
 	nextState.Liquidity = currentLiquidity
 
-	return nil, amount0, amount1, nextState
+	return &SwapResult{
+		StateUpdate:        nextState,
+		amountCalculated:   &cache.amountCalculated,
+		remainingAmountIn:  &cache.amountRequired,
```
piavgh, line 202: should rename this to describe both amount in/out

pkg/source/algebrav1/pool_simulator_bigint.go
```diff
@@ -0,0 +1,182 @@
+package algebrav1
```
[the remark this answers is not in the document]
NgoKimPhu: the bigint one is just to compare (same as univ3). but yeah we can remove this later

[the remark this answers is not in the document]
NgoKimPhu: (this bigint one is the old code)

pkg/source/algebrav1/pool_simulator_bigint.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (p *PoolSimulatorBigInt) CalcAmountOut(param pool.CalcAmountOutParams) (*pool.CalcAmountOutResult, error) {
+	tokenAmountIn := param.TokenAmountIn
+	tokenOut := param.TokenOut
+	var tokenInIndex = p.GetTokenIndex(tokenAmountIn.Token)
+	var tokenOutIndex = p.GetTokenIndex(tokenOut)
+	var zeroForOne bool
+
+	if tokenInIndex >= 0 && tokenOutIndex >= 0 {
```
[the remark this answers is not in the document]
NgoKimPhu: it's old code so I'm reluctant to touch it

pkg/source/algebrav1/pool_simulator.go
NgoKimPhu: oh CalcAmountIn also uses gas?

NgoKimPhu: does onchain-price handle RemainingTokenAmountOut?

piavgh: yes, it's an important thing.
If a pool can not serve the whole amount, the onchain-price service will ignore that pool (to avoid partially filled result)

pkg/source/algebrav1/math.go
NgoKimPhu: actually amountRequired can be either amountIn for CalcAmountOut, or amountOut for CalcAmountIn.
and amountCalculated can be either amountOut for CalcAmountOut, or amountIn for CalcAmountIn.

NgoKimPhu: ah I see, yeah true

pkg/source/algebrav1/pool_simulator.go
NgoKimPhu: fixed

PR #622, opened 2024-11-29
refactor: bebop to use float64

NgoKimPhu:


pkg/liquidity-source/bebop/pool_simulator_test.go
```diff
@@ -71,12 +71,12 @@ func TestPoolSimulator_GetAmountOut(t *testing.T) {
...
 			name:        "it should return error when swap higher than total level", // Total level ~166kMATIC
 			amountIn:    big.NewInt(200_000_000000),
-			expectedErr: ErrAmountInIsGreaterThanHighestPriceLevel,
+			expectedErr: ErrInsufficientLiquidity,
 		},
 		{
 			name:              "it should return correct amountOut when swap in levels",
 			amountIn:          big.NewInt(3_000_000),
-			expectedAmountOut: bigIntFromString("3282719618942083072"),
+			expectedAmountOut: bigIntFromString("3282719618942082560"),
```
lehainam-dev, line 79: Value changes due to float64 precision?

pkg/liquidity-source/bebop/pool_simulator.go
```diff
@@ -200,71 +148,37 @@ func (p *PoolSimulator) swap(amountIn *big.Int, baseToken, quoteToken entity.Poo
...
-
-		if amountLeft.Cmp(zeroBF) == 0 || currentLevelIdx == len(priceLevels) {
-			// We don't skip the used price levels, but just reset its quote to zero.
-			break
+			amountIn -= quote
 		}
 	}
 
-	return priceLevels
+	return nil
```
lehainam-dev, line 183: Why do we return nil here instead of `priceLevels`?
Should we update `updatePriceLevelsState` to not return value?

pkg/liquidity-source/bebop/pool_simulator_test.go
NgoKimPhu: Yes, but it is small enough and probably much smaller than difference between get route and rfq quote

pkg/liquidity-source/bebop/pool_simulator.go
NgoKimPhu: the loops should not reach here; if the code reaches here, it means all levels are already used

PR #625, opened 2024-12-01
feat: Integrate Algebra-Integral v1.2

sunspirit99:


pkg/liquidity-source/algebra/integral/math.go
```diff
@@ -0,0 +1,72 @@
...
+
+	remainder := new(big.Int).Mod(x, y)
+	if remainder.Sign() > 0 {
+		quotient.Add(quotient, bignumber.One)
+	}
+
+	return quotient, nil
+}
+
+func mulDiv(a, b, denominator *big.Int) (*big.Int, error) {
```
NgoKimPhu, line 25: if possible, let's use the lib function already available

NgoKimPhu (approved):

PR #639, opened 2024-12-09
fix: HasReserves should return true for single-sided pools

NgoKimPhu:


piavgh (approved):

piavgh:
CI test didn't pass, please check

PR #643, opened 2024-12-11
refactor: update logic lo

hoanguyenkh:
## Why did we need it?

For new logic fee merchanism we add a new field IsTakerAssetFee in the order payload,
New logic is:

- if IsTakerAssetFee is true then take fee from takerAsset
- if IsTakerAssetFee is false then take fee from makerAsset


pkg/source/limitorder/pool_simulator.go
```diff
@@ -264,10 +264,18 @@ func (p *PoolSimulator) calcAmountWithSwapInfo(swapSide SwapSide, tokenAmountIn
...
 
-			feeAmountWeiByOrder := p.calcFeeAmountPerOrder(order, filledMakingAmountWei)
-			totalFeeAmountWei = new(big.Int).Add(totalFeeAmountWei, feeAmountWeiByOrder)
-			actualAmountOut := new(big.Int).Sub(filledMakingAmountWei, feeAmountWeiByOrder)
+			var feeAmountWeiByOrder *big.Int
+			var actualAmountOut *big.Int
+			if order.IsTakerAssetFee {
+				// Calculate fee in taker asset
+				feeAmountWeiByOrder = p.calcFeeAmountPerOrder(order, remainingTakingAmountWei)
+				actualAmountOut = new(big.Int).Sub(remainingTakingAmountWei, feeAmountWeiByOrder)
```
hieuvo, line 272: I think actualAmountOut should be calculated from makingAmount rather than takingAMount

it4rb (approved):

PR #649, opened 2024-12-12
update(virtual-fun): Adjust the estimated gas if pool triggers logic to deploy Uni pool

sunspirit99:


pkg/liquidity-source/virtual-fun/pool_simulator.go
```diff
@@ -150,10 +153,15 @@ func (s *PoolSimulator) CalcAmountOut(param poolpkg.CalcAmountOutParams) (*poolp
 		isBuy = true
 	}
 
+	gas := s.gas.Swap
+	if newReserveA.Cmp(s.gradThreshold) <= 0 {
```
NgoKimPhu, line 157: double check `<=`

sunspirit99: the condition is correct after checking

NgoKimPhu (approved):

PR #663, opened 2024-12-22
fix/algebra integral concurrent map access

NgoKimPhu:
- **fix: algebra integral concurrent map access**
- **refactor: simplify lock logic to Once logic**


pkg/liquidity-source/algebra/integral/type.go
```diff
@@ -133,22 +121,45 @@ type TimepointRPC struct {
...
 }
 
-// we won't update the state when calculating amountOut, return this struct instead
+type SlidingFeeConfig struct {
+	ZeroToOneFeeFactor *uint256.Int `json:"0to1fF,omitempty"`
+	OneToZeroFeeFactor *uint256.Int `json:"1to0fF,omitempty"`
+}
+
+type StaticExtra struct {
+	UseBasePluginV2 bool `json:"2"`
```
sunspirit99, line 159: do we need a more descriptive tag?

pkg/liquidity-source/algebra/integral/plugin.go
```diff
@@ -537,10 +480,10 @@ func getOutputTokenDelta10(to, from, liquidity *uint256.Int) (*uint256.Int, erro
 }
 
 func getToken0Delta(priceLower, priceUpper, liquidity *uint256.Int, roundUp bool) (*uint256.Int, error) {
-	priceDelta := new(uint256.Int).Sub(priceUpper, priceLower)
-	if priceDelta.Cmp(priceUpper) >= 0 {
-		return nil, errors.New("price delta must be smaller than price upper")
+	if priceUpper.Cmp(priceLower) < 0 {
+		return nil, errors.New("price upper must not be less than price lower")
 	}
```
sunspirit99, line 485: I think we should revert to the old calculation:  
`c = a - b`.  
Check `c >= b` ⇔ `a - b >= b` ⇔ `a >= 2b` (we still need a big.Int multiplication here to ensure the condition is checked correctly)

We can stick to the old approach of performing a big.Int subtraction first and then comparing. The total of big.Int calculations remains the same

sunspirit99, line 485:
```suggestion
	if priceUpper.Cmp(priceLower) < 0 || priceLower.Sign() == 0 {
		return nil, errors.New("price upper must not be less than price lower")
	}
```

sunspirit99: I just add to ensure it is equivalent to the prev condition, the logic outside may have already ensured that `priceLower != 0`

sunspirit99:
LGTM overall, let me try get some routes

pkg/liquidity-source/algebra/integral/math.go
```diff
@@ -1,26 +1,22 @@
...
 	v3Utils "github.com/KyberNetwork/uniswapv3-sdk-uint256/utils"
 	"github.com/holiman/uint256"
 )
 
-func unsafeDivRoundingUp(x, y *uint256.Int) (*uint256.Int, error) {
-	if y.Sign() == 0 {
-		return nil, errors.New("division by zero")
+func unsafeDivRoundingUp(x, y *uint256.Int) *uint256.Int {
+	if y.IsZero() {
+		panic("division by zero")
```
sunspirit99, line 11: I have the same concern with Ms.Chau about the panic

NgoKimPhu: we already checked for 0 before this. anyway removed this panic

pkg/liquidity-source/algebra/integral/plugin.go
NgoKimPhu: it's already checked in `calculateSwap`. btw it's better to compare with MIN_SQRT_RATIO

pkg/liquidity-source/algebra/integral/type.go
NgoKimPhu: changed to `pluginV2`

PR #669, opened 2024-12-26
feat: multiple thegraph api keys

minhnhathoang:


pkg/liquidity-source/algebra/integral/pool_list_updater.go
```diff
@@ -3,49 +3,46 @@ package integral
...
-	graphqlClient *graphql.Client
+	config           *Config
+	graphqlClient    *graphqlpkg.Client
+	graphqlClientCfg *graphqlpkg.Config
 }
 
 func NewPoolsListUpdater(
 	cfg *Config,
+	graphqlClient *graphqlpkg.Client,
+	graphqlClientCfg *graphqlpkg.Config,
```
sunspirit99, line 27: I don't see this `graphqlClientCfg` being used anywhere in this file

pkg/liquidity-source/algebra/integral/pool_list_updater.go
```diff
@@ -3,49 +3,46 @@ package integral
...
 
-	req := graphql.NewRequest(getPoolsListQuery(allowSubgraphError, lastCreatedAtTimestamp, lastPoolIds, first, skip))
+	req := graphqlpkg.NewRequest(getPoolsListQuery(allowSubgraphError, lastCreatedAtTimestamp, lastPoolIds, first, skip))
 
 	var response struct {
 		Pools []SubgraphPool `json:"pools"`
 	}
 
-	if err := d.graphqlClient.Run(ctx, req, &response); err != nil {
+	if err, _ := d.graphqlClient.Run(ctx, req, &response); err != nil {
```
sunspirit99, line 45: I think we're more familiar with convention of returning the error at the last position  image

pkg/util/graphql/graphql.go
```diff
@@ -0,0 +1,333 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"mime/multipart"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+type IClient interface {
```
sunspirit99, line 17: Is this interface complete with all the required methods ?

pkg/liquidity-source/algebra/integral/pool_list_updater.go
minhnhathoang: oh, `graphqlClient` is initialized with config from pool-service. I removed redundant code

pkg/util/graphql/graphql.go
```diff
@@ -0,0 +1,362 @@
...
+	if c.httpClient == nil {
+		c.httpClient = http.DefaultClient
+	}
+	chainClientInterceptors(c)
+	return c
+}
+
+func (c *Client) logf(format string, args ...interface{}) {
+	c.Log(fmt.Sprintf(format, args...))
+}
```
NgoKimPhu, line 56: the `Log` field should have the same api of `(format string, args ...interface{})` so that we don't waste time calling `fmt.Sprintf` if the Log is a no-op

pkg/util/graphql/graphql.go
```diff
@@ -0,0 +1,362 @@
+package graphql
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
```
NgoKimPhu, line 6: use goccy?

pkg/util/graphql/graphql.go
minhnhathoang: yes, I only added custom logic around Client.Invoke() in this version

NgoKimPhu (approved):

PR #681, opened 2025-01-02
ft: optimize maverick-v1

NgoKimPhu:
* Use uint256
* Use primitive int when possible


sunspirit99:
LGTM , still clean as ever

sunspirit99 (approved):

PR #693, opened 2025-01-09
feat : Expand Curve(two-crypto) on sonic

sunspirit99:


pkg/liquidity-source/curve/shared/constant.go
```diff
@@ -33,10 +33,15 @@ const (
...
+	CurveAddressProvider_Sonic = "0x87fe17697d0f14a222e8bef386a0860ecffdd617"
+
 	poolMethodGamma           = "gamma"
 	poolMethodUnderlyingCoins = "underlying_coins"
 
 	addressProviderMethodGetAddress = "get_address"
 
 	getPoolsEndpoint = "/api/getPools/%s/%s" // <chain>/<registry>
+
+	getPoolsEndpoint_Sonic = "/v1/getPools/%s/%s" // <chain>/<registry>
```
NgoKimPhu, line 46: I think `api` and `v1` both works, no need for extra logic here

NgoKimPhu: use `v1`

pkg/source/curve/pools/sonic.json
```diff
@@ -0,0 +1,50 @@
+[
```
NgoKimPhu, line 1: we don't need this file right?

sunspirit99: we still need this

sunspirit99: for old version in pkg/source for backward compatible

NgoKimPhu (approved):

PR #716, opened 2025-02-10
ft/pool factory

NgoKimPhu:
- **ft: pool simulator factory**
- **ft: improve factory test**
- **ft: pool lister and tracker factory**


ChauNguyen-K (approved):

pkg/source/pool/list/factory.go
```diff
@@ -0,0 +1,114 @@
...
+// RegisterFactoryCEG registers a factory function for a pool lister with config, ethrpcClient and graphqlClient
+func RegisterFactoryCEG[C any, P IPoolsLister](poolType string, factory func(*C, *ethrpc.Client, *graphqlpkg.Client) P) bool {
+	return RegisterFactory(poolType, func(params PoolsListerParams[C]) (IPoolsLister, error) {
+		return factory(params.Cfg, params.EthrpcClient, params.GraphqlClient), nil
+	})
+}
+
+// RegisterFactoryE registers a factory function for a pool lister with ethrpcClient
+func RegisterFactoryE[P IPoolsLister](poolType string, factory func(*ethrpc.Client) P) bool {
+	return RegisterFactory(poolType, func(params PoolsListerParams[[0]any]) (IPoolsLister, error) {
```
lehainam-dev, line 98: I'm curious `[0]any` vs `any` differences. Can you explain to me more 🙇?
An article/link/keyword is enough.

lehainam-dev (approved):

pkg/source/pool/list/factory.go
NgoKimPhu: It's very similar to `struct{}` in that [its Sizeof is 0](https://go.dev/play/p/I9pHls8ZjIZ).
I copied the usage from here: https://github.com/golang/go/blob/master/src/sync/atomic/type.go#L47

The difference is that in the atomic example, `[0]*T` retains the type information of `T` compared to `struct{}`.
In our case actually there's no difference, using `struct{}` is also fine.

NgoKimPhu: oh, actually `[0]any` will fail for the json marshal+unmarshal part... gonna update to use `struct{}` instead

PR #730, opened 2025-02-17
Fix some issue in Overnight-usdp, Ramses-v2 & Curve-xCrypto-Ng

sunspirit99:


pkg/liquidity-source/curve/shared/pools_list_updater.go
```diff
@@ -26,7 +26,7 @@ type (
 	}
 
 	Config struct {
-		DexID       string              `mapstructure:"dex_id" json:"dex_id,omitempty"`
+		DexID       string              `mapstructure:"dexID" json:"dexID,omitempty"`
```
NgoKimPhu, line 29: Can we use the same casing style for the other fields as well?

sunspirit99: So we need to update config in pool service too

sunspirit99: i think it's better when do it on the next PR :)

PR #738, opened 2025-02-23
get dependencies for wombat

tien7668:


pkg/source/wombat/pool_tracker.go
```diff
@@ -192,12 +193,13 @@ func (d *PoolTracker) getNewPoolState(
...
-		StartCovRatio: startCovRatio,
-		EndCovRatio:   endcovRatio,
-		AssetMap:      assetMap,
+		Paused:             paused,
+		HaircutRate:        haircutRate,
+		AmpFactor:          ampFactor,
+		StartCovRatio:      startCovRatio,
+		EndCovRatio:        endcovRatio,
+		AssetMap:           assetMap,
+		DependenciesStored: true,
```
NgoKimPhu, line 202: Make this a common struct to be embedded? Use short but unique json tag

NgoKimPhu: But why do we need this?

tien7668: It's for saving dependencies to global for current pool running(default false).
Once it updated to true, it will skip that saving part

NgoKimPhu (approved):

PR #746, opened 2025-03-04
EX-2949: integrate panda-fun

lehainam-dev:


pkg/liquidity-source/pandafun/constant.go
```diff
@@ -0,0 +1,8 @@
+package pandafun
+
+import "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/bignumber"
+
+const defaultGas = 200000 // TODO: Update this
```
NgoKimPhu, line 5: 130k?
image

pkg/liquidity-source/pandafun/pool_simulator.go
```diff
@@ -0,0 +1,200 @@
...
+		Fee: &pool.TokenAmount{
+			Token:  p.Info.Tokens[0],
+			Amount: fee,
+		},
+		Gas: defaultGas,
+	}, nil
+}
+
+func (p *PoolSimulator) UpdateBalance(params pool.UpdateBalanceParams) {
+	// Since we don't need sqrtP in calculation, we only need to update new reserves
```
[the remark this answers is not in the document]
lehainam-dev: @NgoKimPhu we do calculate a `sqrtPNew`, but the calculation doesn't depend on the previous `sqrtP` values. So we don't need to save that.

It depends on `sqrtPa` and `sqrtPb`, which I don't see any mutation of those values in buy/sell tokens.

pkg/liquidity-source/pandafun/constant.go
lehainam-dev: I keep making the same mistake :(

NgoKimPhu (approved):

PR #747, opened 2025-03-06
Ft/balancer meta

sunspirit99:


pkg/liquidity-source/balancer-v2/weighted/pool_tracker.go
```diff
@@ -197,6 +227,84 @@ func (t *PoolTracker) queryRPC(
...
+
+	req := t.ethrpcClient.R().SetContext(ctx)
+	for i, token := range tokens {
+		req.AddCall(&ethrpc.Call{
+			ABI:    poolABI,
+			Target: token.Address,
+			Method: poolMethodGetPoolId,
+			Params: []any{},
+		}, []any{&basePoolIds[i]})
+	}
```
NgoKimPhu, line 241: maybe we should adapt to https://test-api-v3.balancer.fi/. this api gives correct result for tokens as nestedPools

pkg/liquidity-source/balancer-v2/weighted/pool_simulator.go
```diff
@@ -493,3 +759,89 @@ func _upscaleArray(poolTypeVer int, balances []*big.Int, scalingFactors []*uint2
...
+	return lo.Keys(tokenSet)
+}
+
+func (t *PoolSimulator) GetBasePool() pool.IPoolSimulator {
+	for _, basePool := range t.basePools {
+		return basePool
+	}
+
+	return nil
+}
```
NgoKimPhu, line 841: hmm, need to return all?

pkg/liquidity-source/balancer-v2/shared/subgraph.go
```diff
@@ -16,6 +16,11 @@ type SubgraphPool struct {
 		Address  string `json:"address"`
 		Decimals int    `json:"decimals"`
 		Weight   string `json:"weight"`
+		Token    struct {
+			Pool struct {
+				ID string `json:"id"`
```
NgoKimPhu, line 21: prefer address over id

pkg/liquidity-source/balancer-v2/stable/pool_tracker.go
```diff
@@ -98,6 +98,37 @@ func (t *PoolTracker) getNewPoolState(
 	}
 	scalingFactors := oldExtra.ScalingFactors
 
+	if !staticExtra.BasePoolScanned {
```
NgoKimPhu, line 101: optionally can just rescan all pools, no need for this adhoc logic. if already implemented then also fine

pkg/liquidity-source/balancer-v2/stable/pool_simulator.go
```diff
@@ -150,17 +232,316 @@ func (s *PoolSimulator) CalcAmountOut(params pool.CalcAmountOutParams) (*pool.Ca
...
+		PoolId:        basePool.GetPoolId(),
+		Pool:          basePool.GetAddress(),
+		TokenIn:       tokenIn,
+		TokenOut:      bptToken,
+		AmountIn:      amountIn.ToBig(),
+		AmountOut:     bptAmount.ToBig(),
+		JoinExitIndex: joinIndex,
+	})
+
+	amountOut, err := s.OnSwap(tokenOut, bptToken, bptAmount)
```
minhnhathoang, line 360: Is the token order flipped?
```suggestion
	amountOut, err := s.OnSwap(bptToken, tokenOut, bptAmount)
```

pkg/liquidity-source/balancer-v2/stable/pool_simulator.go
```diff
@@ -150,17 +232,316 @@ func (s *PoolSimulator) CalcAmountOut(params pool.CalcAmountOutParams) (*pool.Ca
...
+		amountOut *uint256.Int
+		exitIndex *big.Int
+	)
+
+	switch basePool.GetType() {
+	case composablestable.DexType:
+		amountOut, err = basePool.OnSwap(bptToken, tokenOut, bptAmount)
+	default:
+		amountOut, err = basePool.OnExit(tokenOut, bptAmount)
+		exitIndex = shared.PackJoinExitIndex(shared.PoolExit, 1)
```
minhnhathoang, line 411:
```suggestion
		exitIndex = shared.PackJoinExitIndex(shared.PoolExit, len(hops))
```
Should use len(hops) instead of hardcoded value

pkg/liquidity-source/balancer-v2/shared/constant.go
```diff
@@ -1,5 +1,15 @@
 package shared
 
 const (
+	ZeroAddress              = "0x0000000000000000000000000000000000000000000000000000000000000000"
```
minhnhathoang, line 4: Can use eth.ZeroAddress

pkg/liquidity-source/balancer-v2/math/fixed_point.go
```diff
@@ -128,6 +128,67 @@ func (l *fixedPoint) PowUp(x *uint256.Int, y *uint256.Int) (*uint256.Int, error)
...
+	maxError, err := l.Add(mulUpRawAndMaxPow, number.Number_1)
+	if err != nil {
+		return nil, err
+	}
+
+	if raw.Lt(maxError) {
+		return number.Zero, nil
+	}
+
+	return l.Sub(raw, maxError)
```
minhnhathoang, line 166:
```suggestion
	return l.PowDownV1(x, y)
```

Why do we have both `PowDown` and `PowDownV1`? 
I think we should either keep only one version if there's no significant difference, or reuse `PowDownV1` here to avoid code duplication.

pkg/liquidity-source/balancer-v2/shared/constant.go
NgoKimPhu: btwq this looks like a bytes32, not an address

pkg/liquidity-source/balancer-v2/shared/subgraph.go
```diff
@@ -16,6 +16,11 @@ type SubgraphPool struct {
 		Address  string `json:"address"`
 		Decimals int    `json:"decimals"`
 		Weight   string `json:"weight"`
+		Token    struct {
+			Pool struct {
+				Address string `json:"address"`
+			} `json:"pool"`
+		} `json:"token"`
 	} `json:"tokens"`
 }
```
NgoKimPhu, line 26: let's migrate to use `https://test-api-v3.balancer.fi/` ?

NgoKimPhu (approved):

PR #750, opened 2025-03-06
Ekubo Integration

die-herdplatte:
For backend integration, [Ekubo's router](https://github.com/paraswap/paraswap-dex-lib/blob/aa3c89421610914d5bc45de49db597769eed7468/src/abi/ekubo/router.json) can be used.
Exact out swaps are encoded by negating the input amount and specifying the token to be the output token.
Besides `swap`, the router also provides a `multihopSwap` and `multiMultihopSwap` functions to connect multiple swaps without the need to transfer any tokens.
`sqrtRatioLimit` is the limit price, the min and max constants can be found [here](https://github.com/paraswap/paraswap-dex-lib/blob/aa3c89421610914d5bc45de49db597769eed7468/src/dex/ekubo/pools/math/price.ts).
The `SwapInfo` of this integration contains the last parameter of a swap call, `skipAhead`.

The included tests in `pool_tracker_test.go` and `pools_list_updater.go` require setting the `ETH_RPC_URL` environment variable to a Sepolia RPC endpoint.

 Ekubo V2 contracts are only deployed on Sepolia for now, contract addresses still have to be updated and tests changed once the mainnet contracts are deployed.


pkg/source/ekubo/pool_simulator.go
```diff
@@ -0,0 +1,123 @@
...
+		RemainingTokenAmountOut: &pool.TokenAmount{
+			Token:  params.TokenAmountOut.Token,
+			Amount: new(big.Int).Add(params.TokenAmountOut.Amount, quote.ConsumedAmount),
+		},
+		Gas:      quote.Gas,
+		SwapInfo: quote.SkipAhead,
+	}, nil
+}
+
+func (p *PoolSimulator) UpdateBalance(_ pool.UpdateBalanceParams) {}
```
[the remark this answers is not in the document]
die-herdplatte: Got it, I implemented it.
Still left the `SkipAhead` property in the `SwapInfo` because I thought that you may need it for the backend integration.

pkg/source/ekubo/pool_tracker.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	n.Rsh(n, 32)
+
+	sqrtRatioAfterCompact := new(big.Int).And(
+		n,
+		math.U96Max,
+	)
+	n.Rsh(n, 96)
+
+	poolState.SqrtRatio = math.FloatSqrtRatioToFixed(sqrtRatioAfterCompact)
```
minhnhathoang, line 149:
```suggestion
    poolState.FixedSqrtRatio = math.FloatSqrtRatioToFixed(sqrtRatioAfterCompact)
```
Rename the variable so it’s clear which sqrtRatio is fixed and which isn’t. Also, ensure that `quoting.PoolState.sqrtRatio` is always calculated using one format. At present, the field takes non-fixed data from QuoteData and converts it to fixed when handling events.

die-herdplatte: > Rename the variable so it’s clear which sqrtRatio is fixed and which isn’t.

I now suffixed the `QuoteData` property with `Fixed` and kept the state property as-is since that follows the naming conventions of Ekubo's Rust SDK.

> Also, ensure that quoting.PoolState.sqrtRatio is always calculated using one format. At present, the field takes non-fixed data from QuoteData and converts it to fixed when handling events.

Good catch, thanks!

NgoKimPhu (approved):

PR #788, opened 2025-04-03
ft/kem

NgoKimPhu:
- **refactor: pool rfq**
- **ft: integrate kem**


pkg/liquidity-source/syncswapv2/shared/pools_list_updater.go
```diff
@@ -2,16 +2,17 @@ package syncswapv2shared
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/goccy/go-json"
+
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/entity"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/liquidity-source/syncswapv2"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/syncswap"
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/util"
-	"github.com/KyberNetwork/logger"
-	"github.com/ethereum/go-ethereum/common"
 )
```
lehainam-dev, line 16: Can you share your VSCode setup? I also want my config to be the same in package arrangement.

pkg/liquidity-source/uniswap-v4/hooks.go
```diff
@@ -38,3 +42,27 @@ func HasSwapPermissions(address common.Address) bool {
 	// This implicitly encapsulates swap delta permissions
 	return hasPermission(address, BeforeSwap) || hasPermission(address, AfterSwap)
 }
+
+type Hook interface {
+	GetExchange() string
+	RFQ(context.Context, pool.RFQParams, *PoolMetaInfo, *pool.RFQResult) (any, error)
```
lehainam-dev, line 48: It seems like RFQ is only supported in KEM, and not supported in other hooks. Should we add such method into the interface?

pkg/source/pool/rfq.go
```diff
@@ -3,35 +3,43 @@ package pool
...
+	RFQRecipient string              // RFQ fund recipient (executor/next pool/recipient)
+	Source       string              // source client
+	TokenIn      string              // address of token swap from
+	TokenOut     string              // address of token swap to
+	SwapAmount   *big.Int            // amount of TokenIn to swap
+	AmountOut    *big.Int            // amount of TokenOut received
+	Slippage     int64               // tolerance (in bps) for RFQs that also aggregate dexes
+	PoolExtra    any                 // extra pool metadata
+	SwapInfo     any                 // swap info of the RFQ swap
+	FeeInfo      any                 // generic fee info
```
lehainam-dev, line 26: Do you have any extension to support the struct field arrangement? (It seems like your struct is more optimized regarding alignment padding).

pkg/liquidity-source/syncswapv2/shared/pools_list_updater.go
minhnhathoang: You can use the formatter on Goland mentioned [here](https://github.com/KyberNetwork/pool-service/pull/946#discussion_r1903973165) , I think VSCode can be configured similarly.

NgoKimPhu: My GoLand set-up: 
image

But I usually run this command before committing, it's more consistent because it merges import groups beforehand:

```bash
(){sed -i '/^import (/,/^)/{/^$/d}' $*; goimports -local `grep -oP 'module \K.+' go.mod` -w $*} `git d origin/main --name-status | grep -v '^D' | awk '{print $NF}' | grep '.go$'`
```

pkg/liquidity-source/uniswap-v4/hooks.go
NgoKimPhu: the `BaseHook` will just return nil, meaning no-op. I think no-op base methods are better than specific method per hook? in the future if we have other similar pools it will be simpler to expand (and actually we do have some similar pools to integrate like this in the future)

pkg/source/pool/rfq.go
NgoKimPhu: oh no haha, just did it manually. golang compiler nowadays should be smart enough to do this alignment already, no need to optimize manually. I just here put fields related to each other closer to each other, that's it (for OCD reasons, lol)

PR #799, opened 2025-04-18
feat(*): expose NextStateSqrtRatioX96 field in swapInfo

linhbkhn95:
Currently, we need `NextStateSqrtRatioX96` at some internal services. But it didn't return in router API due to this field is private.
## Why did we need it?

- Set `NextStateSqrtRatioX96` as public field

### DIFF
--- a/pkg/source/uniswapv3/type.go
+++ b/pkg/source/uniswapv3/type.go
@@ -18,7 +18,7 @@ type Gas struct {
 }
 
 type SwapInfo struct {
-	nextStateSqrtRatioX96 *v3Utils.Uint160
+	NextStateSqrtRatioX96 *v3Utils.Uint160 `json:"nSqrtRx96"`
 	nextStateLiquidity    *v3Utils.Uint128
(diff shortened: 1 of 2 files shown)


pkg/source/uniswapv3/type.go
```diff
@@ -18,7 +18,7 @@ type Gas struct {
 }
 
 type SwapInfo struct {
-	nextStateSqrtRatioX96 *v3Utils.Uint160
+	NextStateSqrtRatioX96 *v3Utils.Uint160 `json:"nextStateSqrtRatioX96"`
```
NgoKimPhu, line 21:
```suggestion
	NextStateSqrtRatioX96 *v3Utils.Uint160 `json:"nSqrtRx96"`
```

hieuvo, line 21: should we export all 3 fields, instead of only NextStateSqrtRatioX96?

NgoKimPhu (approved):

PR #804, opened 2025-04-23
fix: Handle bebop call rfq fail

chrisngyn:
## Why did we need it?

When an error occurs when calling rfq, bebop still return error response with http status is 200. So I update to handle it.


thanhpp (approved):

lehainam-dev (approved):
LGTM

PR #819, opened 2025-05-05
fix concurrent RW issue on woofi-v3

sunspirit99:

### DIFF
--- a/pkg/liquidity-source/woofi-v21/pool_simulator.go
+++ b/pkg/liquidity-source/woofi-v21/pool_simulator.go
@@ -673,3 +673,36 @@ func (s *PoolSimulator) updateBalanceSwapBaseToBase(params pool.UpdateBalancePar
 		WoFeasible: stateOut.WoFeasible,
 	}
 }
+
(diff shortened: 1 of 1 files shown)


pkg/liquidity-source/woofi-v21/pool_simulator.go
```diff
@@ -674,11 +673,3 @@ func (s *PoolSimulator) updateBalanceSwapBaseToBase(params pool.UpdateBalancePar
 		WoFeasible: stateOut.WoFeasible,
 	}
 }
-
-func (p *PoolSimulator) CloneState() pool.IPoolSimulator {
```
sunspirit99, line 678: clean to fall back to deep copy

NgoKimPhu: should we add clone for `wooracle.States` instead?

sunspirit99: it also okay now

sunspirit99: updated at latest commit

NgoKimPhu (approved):

PR #831, opened 2025-05-13
feat: add custom approval address

minhnhathoang:
## Why did we need it?
[EX-3205](https://team-kyber.atlassian.net/browse/EX-3205)
[Note](https://www.notion.so/kybernetwork/Improve-pool-approval-tracking-1f226751887e80be8991d85b44574b50?pvs=4)

## Screenshots (if appropriate):

[EX-3205]: https://team-kyber.atlassian.net/browse/EX-3205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/source/pool/pool.go
```diff
@@ -38,6 +38,10 @@ func (t *Pool) CalculateLimit() map[string]*big.Int {
 	return nil
 }
 
+func (t *Pool) GetApprovalAddress(tokenIn, tokenOut string) string {
+	return ""
+}
```
lehainam-dev, line 43: Should we default to the return pool address instead?

NgoKimPhu (approved):

pkg/source/pool/pool.go
minhnhathoang: I return an empty string by default so pools that don't call approve don't need to override this method. If the pool doesn't call approve max or approval address is pool itself, it will be determined by function selector in the router service.

PR #892, opened 2025-06-18
feat: add pancake-infinity dynamic fee and brevis hooks

minhnhathoang:


pkg/liquidity-source/pancake-infinity/bin/pool_tracker.go
```diff
@@ -243,6 +268,25 @@ func (t *PoolTracker) getBinsFromSubgraph(ctx context.Context, poolAddress strin
...
+			}).Errorf("failed to get fee for %s hook %s", hook.GetExchange(), hookAddress.Hex())
+			return shared.MAX_FEE_PIPS
+		}
+		return uint32(fee.Uint64())
+
+	default:
+		return shared.MAX_FEE_PIPS
+	}
+}
+
```
NgoKimPhu, line 289: should we make this an interface function of the hooks? instead of switch case

pkg/valueobject/exchange.go
```diff
@@ -3,814 +3,820 @@ package valueobject
...
+	ExchangeOvernightUsdp               = "overnight-usdp"
+	ExchangeOwlSwapV3                   = "owlswap-v3"
+	ExchangePaintSwap                   = "paintswap"
+	ExchangePancake                     = "pancake"
+	ExchangePancakeInfinityBin          = "pancake-infinity-bin"
+	ExchangePancakeInfinityBinBrevis    = "pancake-infinity-bin-brevis"
+	ExchangePancakeInfinityBinFairflow  = "pancake-infinity-bin-fairflow"
+	ExchangePancakeInfinityCL           = "pancake-infinity-cl"
+	ExchangePancakeInfinityCLBrevis     = "pancake-infinity-cl-brevis"
+	ExchangePancakeInfinityCLDynamicFee = "pancake-infinity-cl-dynamic-fee"
```
NgoKimPhu, line 248: what about
```suggestion
	ExchangePancakeInfinityCLDynamic = "pancake-infinity-cl-dynamic"
```

pkg/liquidity-source/pancake-infinity/bin/pool_tracker.go
```diff
@@ -122,19 +126,40 @@ func (t *PoolTracker) GetNewPoolState(
...
+		l.WithFields(logger.Fields{
+			"error": err,
+		}).Error("failed to unmarshal static extra")
+	} else {
+		hookAddress = staticExtra.HooksAddress
+	}
+
+	lpFee := staticExtra.Fee
+	if staticExtra.IsDynamicFee {
+		lpFee = t.GetDynamicFee(ctx, hookAddress)
```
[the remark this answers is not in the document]
minhnhathoang: For Brevis hooks, we’re using [origFee](https://github.com/brevis-network/pancake-tokenholding-hook/blob/06c3ea757d747cce020ecce98e1ee7918ac96874/src/VipDiscountMap.sol#L13) as default max fee (no discount), and it is currently immutable for now. For common dynamic fee hooks, the fee can be updated via the [DynamicLPFeeUpdated](https://github.com/pancakeswap/infinity-core/blob/870eb15/src/interfaces/IPoolManager.sol#L22) event, I will update PR accordingly. Thanks for pointing it out!

pkg/liquidity-source/pancake-infinity/bin/pool_tracker.go
minhnhathoang: Updated

PR #893, opened 2025-06-20
eulerswap: use poolsSlice to fetch new pools

sunspirit99:

### DIFF
--- a/pkg/liquidity-source/uniswap-v4/hooks/euler/pool_list_updater.go
+++ b/pkg/liquidity-source/uniswap-v4/hooks/euler/pool_list_updater.go
@@ -2,6 +2,7 @@ package euler
 
 import (
 	"context"
+	"math/big"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/liquidity-source/uniswap-v4/hooks/euler/pool_list_updater.go
```diff
@@ -66,7 +78,7 @@ func (u *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte
 		return nil, metadataBytes, err
 	}
 
-	newMetadataBytes, err := u.newMetadata(offset + len(pools))
+	newMetadataBytes, err := u.newMetadata(offset + batchSize)
```
NgoKimPhu, line 81: Update offset in case of uninstall

NgoKimPhu (approved):

PR #908, opened 2025-06-26
chore: log block number in FetchRPCData

minhnhathoang:


pkg/liquidity-source/pancake-infinity/bin/pool_tracker.go
```diff
@@ -78,6 +77,11 @@ func (t *PoolTracker) FetchRPCData(ctx context.Context, p *entity.Pool, blockNum
 	protocolFee := result.Slot0.ProtocolFee
 	result.SwapFee = lo.Ternary(protocolFee == 0, uint64(lpFee), uint64(calculateSwapFee(protocolFee, lpFee)))
 
+	logger.WithFields(logger.Fields{
```
sunspirit99, line 80: blockNumber is set in the request why do we need to log it frequently ?

PR #923, opened 2025-07-04
Assign origin instead of sender in Firm request 

sunspirit99:


pkg/source/pool/rfq.go
```diff
@@ -15,6 +15,7 @@ type RFQParams struct {
 	NetworkID    valueobject.ChainID // blockchain network id
 	RequestID    string              // request id from getRoute
 	PoolID       string              // pool id
+	Origin       string              // original address
 	Sender       string              // swap tx origin
```
NgoKimPhu, line 19:
```suggestion
	Origin       string              // swap tx tx.origin
	Sender       string              // swap tx msg.sender
```

NgoKimPhu (approved):

PR #925, opened 2025-07-07
EX-3416: integrate pool-party

lehainam-dev:


pkg/liquidity-source/pool-party/pool_simulator.go
```diff
@@ -0,0 +1,107 @@
...
+			Amount: amountOut,
+		},
+		Fee: &pool.TokenAmount{
+			Token:  s.Info.Tokens[0],
+			Amount: nil,
+		},
+		Gas: defaultGas,
+	}, nil
+}
+
```
NgoKimPhu, line 100: implement CloneState. (and use copy-on-write in UpdateBalance)

pkg/liquidity-source/pool-party/pool_tracker.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := g.Wait(); err != nil {
+		l.WithFields(logger.Fields{
+			"error": err,
+		}).Error("failed to fetch pool state")
+		return entity.Pool{}, err
+	}
+
+	extra := Extra{
+		PoolStatus:            subgraphPool.PoolStatus,
+		IsVisible:             subgraphPool.IsVisible,
```
NgoKimPhu, line 112: instead of checking these conditions in simulator, we should check it right from update step (and set reserve to 0) to not waste a slot in the router index

PR #978, opened 2025-08-05
Improve eulerswap pool list updater to avoid missing pool for some case

sunspirit99:

### DIFF
--- a/pkg/util/testutil/calc_amount_out.go
+++ b/pkg/util/testutil/calc_amount_out.go
@@ -16,15 +16,9 @@ func TestCalcAmountOut[TB interface {
 	Run(string, func(TB)) bool
 }](tb TB, poolSim pool.IPoolSimulator, expected map[int]map[int]map[string]string) {
 	tokens := poolSim.GetTokens()
-	for idxIn, expectedByTokenIn := range expected {
-		for idxOut, expectedByTokenOut := range expectedByTokenIn {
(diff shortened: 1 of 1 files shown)


pkg/liquidity-source/euler-swap/pool_list_updater.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 type PoolsListUpdater struct {
-	config       *Config
-	ethrpcClient *ethrpc.Client
+	config            *Config
+	ethrpcClient      *ethrpc.Client
+	latestPoolAddress common.Address
```
NgoKimPhu, line 23: put in metadata instead?

pkg/liquidity-source/euler-swap/pool_list_updater.go
```diff
@@ -58,11 +59,36 @@ func (u *PoolsListUpdater) GetNewPools(ctx context.Context, metadataBytes []byte
...
+			Warn("no pools found")
+
 		return nil, metadataBytes, nil
 	}
 
-	if offset > allPoolsLength {
+	needResetOffset := offset > allPoolsLength
+
+	if offset == allPoolsLength {
+		latestPoolAddress, err := u.getLatestPool(ctx, allPoolsLength)
```
NgoKimPhu, line 73: can we just call initPools with offset=-1?

NgoKimPhu (approved):

PR #1026, opened 2025-09-03
feat: Xpress Protocol

matsakiv:
Added support for Xpress Protocol (on-chain orderbook dex on Sonic):

- [x] pool_list_updater.go
- [x] pool_simulator.go
- [x] pool_tracker.go
- [x] pool_list_updater_test.go
- [x] pool_simulator_test.go
- [x] pool_tracker_test.go

## Notes

In the current implementation, pools are requested from the Xpress Protocol API, but in the near future it is possible to change/add contracts to obtain a list of pools from the blockchain.

The state of the pools is requested from the helper contract, but it is also technically possible to switch to events.

### DIFF
--- a/pkg/liquidity-source/xpress/abi.go
+++ b/pkg/liquidity-source/xpress/abi.go
@@ -1,0 +1,30 @@
+package xpress
+
+import (
+	"bytes"
+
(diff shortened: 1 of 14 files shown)


NgoKimPhu:
continue in #1038

PR #1036, opened 2025-09-09
track paused status

tien7668:


pkg/source/saddle/type.go
```diff
@@ -50,7 +51,8 @@ type Gas struct {
 }
 
 type Meta struct {
-	TokenInIndex  int `json:"tokenInIndex"`
-	TokenOutIndex int `json:"tokenOutIndex"`
-	PoolLength    int `json:"poolLength"`
+	TokenInIndex  int    `json:"tokenInIndex"`
+	TokenOutIndex int    `json:"tokenOutIndex"`
+	PoolLength    int    `json:"poolLength"`
+	LpToken       string `json:"lpToken"`
```
lehainam-dev, line 57: Why do we need this field? I don't see any usage of this field in this PR.

lehainam-dev: So we also need to update the encoding package as well? 😭

PR #1061, opened 2025-09-24
feat: balancer-v3 support approval address for batch swap

lehainam-dev:


pkg/liquidity-source/balancer/v3/shared/type.go
```diff
@@ -15,6 +15,7 @@ type StaticExtra struct {
 
 type Extra struct {
 	HooksConfig                `json:"hook"`
+	ChainID                    int            `json:"chain,omitempty"`
```
NgoKimPhu, line 18:
```suggestion
	ChainID                    int            `json:"-"`
```

NgoKimPhu: but what is this field for?

NgoKimPhu (approved):

PR #1097, opened 2025-10-06
Update(balancer-v3): Check max deposit/withdraw with buffer swap

sunspirit99:


pkg/liquidity-source/balancer/v3/shared/pool_tracker.go
```diff
@@ -1,42 +1,66 @@
...
+		req.AddCall(&ethrpc.Call{
+			ABI:    ERC4626ABI,
+			Target: bufferToken,
+			Method: ERC4626MethodMaxDeposit,
+			Params: []any{VaultAddress},
+		}, []any{&maxDeposits[i]})
+		req.AddCall(&ethrpc.Call{
+			ABI:    ERC4626ABI,
+			Target: bufferToken,
+			Method: ERC4626MethodMaxWithdraw,
```
NgoKimPhu, line 48: should we use max withdraw or max redeem?

sunspirit99: yes, it should be max redeem

NgoKimPhu (approved):

PR #1208, opened 2025-12-11
feat: implement limit order hook with order tracking and state manage…

datluongductuan:
…ment

- Added a new limit order hook with functionality to track pending fill orders and manage their states.
- Introduced data structures for order information and status.
- Enhanced the BeforeSwap method to process pending limit orders before executing swaps.
- Updated hook registration to include new functionalities and ABI integration.


pkg/liquidity-source/pancake/infinity/cl/pool_simulator.go
```diff
@@ -49,8 +50,14 @@ func NewPoolSimulator(entityPool entity.Pool, chainID valueobject.ChainID) (*Poo
 		return nil, shared.ErrUnsupportedHook
 	}
 
-	var allowEmptyTicks bool
+	// modify ticks before new pool simulator, some hooks will need this.
+	// In the the original logic, we should do this logic in BeforeSwap in CalcAmountOut, but using here for simplicity.
+	err := hook.ModifyTicks(context.Background(), extra.ExtraTickU256)
```
lehainam-dev: Seem like we also don't use context inside ModifyTicks. Can we skip this field?

datluongductuan: yeah I removed this ctx from params

PR #1211, opened 2025-12-12
fix: erc4626 and balancer-v3 wrong calc amount in

chiengyn:
## Why did we need it?

the `GetClosestRate` only support for exactIn, but it's used for `CalcAmountIn` too. So I updated it for support both `CalcAmountOut` and `CalcAmountIn`.

Because balancer-v3 use this function too, so I updated it too.

### DIFF
--- a/pkg/liquidity-source/balancer/v3/eclp/pool_simulator_test.go
+++ b/pkg/liquidity-source/balancer/v3/eclp/pool_simulator_test.go
@@ -155,7 +155,7 @@ func TestPoolSimulator_CalcAmountIn(t *testing.T) {
 				Amount: big.NewInt(997094),
 			},
 			tokenIn:          "0x6440f144b7e50d6a8439336510312d2f54beb01d",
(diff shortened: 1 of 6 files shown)


pkg/liquidity-source/balancer/v3/eclp/pool_simulator_test.go
```diff
@@ -155,7 +155,7 @@ func TestPoolSimulator_CalcAmountIn(t *testing.T) {
 				Amount: big.NewInt(997094),
 			},
 			tokenIn:          "0x6440f144b7e50d6a8439336510312d2f54beb01d",
-			expectedAmountIn: "999996963127157068",
+			expectedAmountIn: "1307826060618461310",
```
chiengyn: Is this value make more sense? in CalcAmountOut above, it use `1e18` -> `997094`. So when CalcAmountIn, I think amountIn should be greater than `1e18` right?

PR #1216, opened 2025-12-14
Feat: Add Infinifi Gateway Liquidity Source (iUSD, siUSD, liUSD)

0xth4nh:
## Why did we need it?

We needed to integrate Infinifi protocol into KyberSwap to enable users to efficiently mint / swap into Infinifi's token either through Gateway contracts or soon-to-be-bootstrapped Curve Pool.

## What was implemented?

A complete liquidity source for the Infinifi Gateway contract (`0x3f04b65Ddbd87f9CE0A2e7Eb24d80e7fb87625b5`) supporting all synchronous operations through a virtual pool

### Supported Swap Paths

- **USDC → iUSD** (mint) - 1:1 conversion with decimal scaling (6→18 decimals), Gas: ~550k
- **iUSD → USDC** (redeem) - 1:1 conversion with decimal scaling (18→6 decimals), Gas: ~200k
- **iUSD → siUSD** (stake) - ERC4626 vault conversion using live exchange rate, Gas: ~200k
- **siUSD → iUSD** (unstake) - ERC4626 vault conversion using live exchange rate, Gas: ~200k
- **iUSD → liUSD** (lock) - 6 different liUSD tokens with unwinding epochs (1, 2, 4, 6, 8, 13 months), uses per-bucket exchange rates from LockingController, Gas: ~200k

## How Has This Been Tested?
Through unit tests.

### DIFF
--- a/pkg/liquidity-source/infinifi/gateway/config.go
+++ b/pkg/liquidity-source/infinifi/gateway/config.go
@@ -6,17 +6,6 @@ type Config struct {
 	USDC    string `json:"usdc"`
 	IUSD    string `json:"iusd"`
 	SIUSD   string `json:"siusd"`
-	
-	// Multiple liUSD tokens (one per unwinding epoch)
-	LIUSDTokens []LIUSDToken `json:"liusdTokens"`
-	
 	// LockingController address to fetch liUSD token addresses
 	LockingController string `json:"lockingController"`
 }
-
-type LIUSDToken struct {
-	Address         string `json:"address"`         // LockedPositionToken address
-	UnwindingEpochs uint32 `json:"unwindingEpochs"`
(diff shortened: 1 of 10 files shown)


tien7668:
Hi, About iUSD → liUSD (lock) process: Correct me if I’m wrong that we have multiple lock periods with different rates. Then, if we pick the best rate for users, those liUSD tokens will be locked into one of the configured period, right? 
I’m a bit concerned that users won’t know in advance that the liUSD they receive will be locked, possibly for an undesirably long period.

In addition, from getEnabledBuckets, I got 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13 (more than 6). Is this where we get all lock periods?

Issue #1519, opened 2026-07-07
Add Liquidity Party DEX


timolson:
<img width="541" height="78" alt="Liquidity Party" src="https://github.com/user-attachments/assets/02e2fc85-4f80-4f5e-829e-5c39f927110b" />

#### [Liquidity Party](https://liquidity.party) is a game-theoretic multi-asset AMM
- Game theoretic pricing curve based on the [Logarithmic Market Scoring Rule](https://mason.gmu.edu/~rhanson/mktscore.pdf) (LMSR) instead of Constant Product
- Multi-asset support with pools up to 50 tokens
- Fully open source and audited [source code](https://github.com/Liquidity-Party/lmsr-amm)
- Traditional on-chain noncustodial DEX AMM with factory-generated, verifiable pools that issue LP tokens as ERC20's.

Our pools offer multiple advantages to solvers:
* Full liquidity on rare pairs in a single hop: Our multi-asset pools offer all token combinations with full liquidity, so AAVE/PEPE trades with the same liquidity as AAVE/USDC in a one-hop swap, with no need to route multiple hops through a stablecoin. Save both gas and  pool fees on all your swaps.
* Game-theoretic pricing curve: LMSR offers deeper liquidity / less slippage than Constant Product, giving both better prices for takers and higher yields for LP's.

For more information, see our [About page](https://liquidity.party/about) or the source [README.md](https://github.com/Liquidity-Party/lmsr-amm/blob/main/README.md)

For the production contract deployment addresses, see [liqp-deployments.json](https://github.com/Liquidity-Party/lmsr-amm/blob/main/deployment/liqp-deployments.json) (Etherscan and Sourcify verified contracts, non-upgradable.) Our factory is the [PartyPlanner](https://etherscan.io/address/0x5e9db9fa66aea7f254d4a6783b1a6180c4b8aae3#code). We are currently only on mainnet but plan to expand to L2's.

PR #1520, opened 2026-07-07
feat: add Liquidity Party DEX source

timolson:
logo-light

For an introduction to the protocol, see our accompanying issue https://github.com/KyberNetwork/kyberswap-dex-lib/issues/1519
The related PR for execution is https://github.com/KyberNetwork/ks-dex-adapter-lib/pull/6

# Implementation Notes
- Ported the [ABDK math library](https://github.com/abdk-consulting/abdk-libraries-solidity) to Go
  - Placed under `util/abdkmath64x64` so other adapters can re-use it
  - Uses the _signed_ int256 library
  - Certain ABDK ops rely on unchecked overflow behavior, which we simulated in order to provide code congruity with the original solidity lib
  - tests demonstrate ulp-exactness of the port
- We use events for pool discovery, but pool state is refreshed via RPC unconditionally every block
  - EMA's that affect swaps change every block, even if no event is emitted
  - There will only ever be a small number of pools, because:
    - Pool creation is admin-only
    - Each pool has many assets and supports a large number of pair combinations: one 10-asset pool has 45 pairs and a single 50-asset pool offers 1225 pairs
    - Only two pools at launch
    - Complete state is fetched with a single RPC call per pool
    - Complete state tracking from events would require re-implementation of four additional mutating functions, which are _all significantly more complicated than swap._ The AI estimates this would add  1900–2700 lines of code to the adapter that must all be wei-exact
    - Even with this work, there are rare edge cases that would cause the internal Kyber state to still drift from the chain state (e.g. if there is an unexpected donation/transfer/overpayment of coins to the pool)
    - At launch there should only be two RPC calls per block for our entire protocol, and probably never more than ten, ever.
- Our protocol is fee-on-output but all the other adapters set `Fee.Token` to the input token. We set this to the output token.
- We enforce one swap per route by using the `SingleSwapSourceSet` because:
  - The LMSR math guarantees there is never any intra-pool arbitrage available.
  - All token pair combinations in each pool are available with full liquidity, making chained swaps unnecessary.
- The execution calldata encoder is not in these repos, so it's unclear whether we have any discretion in the encoding. Our swap executor requires token indexes as well as the input token address and follows the encoding format of the Uniswap v2 adapter.

# Testing
- The ABDK port has its own test cases. We generated test case vectors from the original Solidity library and have Go tests to ensure exact compatibility
  - `TestGoldenVectors` exercises every ADBK operation and compares against the Solidity-generated vectors
  - `TestBounds` checks reverts and error cases that could not generate test vectors
- Self-contained simulator tests in `pool_simulator_test.go`
  - Runs against a snapshot derived from our real mainnet test pool without any RPC calls
  - Checks `CalcAmountOut` and `CalcAmountIn` for correctness and consistency
  - Tests error conditions
  - Checks that `GetMetaInfo` returns the correct token indexes (token indexes are necessary for our swap call)
- Mainnet integration test `integration_test.go`
  - Constructs a simulator from a pinned mainnet block that has our live test pool
  - Verifies pool discovery/initialization
  - Tests that simulated amounts are wei-exact matches with "live" execution for both exact-in and exact-out swaps
- Executor test in `ks-dex-adapter-lib` `LiquidityPartyAdapter.t.sol`
  - Tests calldata decoding
  - Checks all six swap combinations with exact in/out amount tests


NgoKimPhu:
merged in 0846ddc1

PR #1584, opened 2026-08-04
feat: integrate Metronome synthetic-asset swap (metronome-swap)

NgoKimPhu:
Adds pkg/liquidity-source/metronome-swap for Metronome's Pool.swap() —
an oracle-rate conversion between synthetic tokens (msETH/msUSD/msBTC/...)
with a static per-ordered-pair fee, no bonding curve and no pooled
reserves. Capacity is bounded by each output token's maxTotalSupply,
checked against the gross (fee+net) minted amount since Pool.swap()
mints the fee to a separate collector in addition to the net amount to
the trader.

CalcAmountOut/UpdateBalance were verified byte-exact against real
on-chain swaps (both historical and live Tenderly vnet execution),
including fee round-half-up rounding (Aave-style WadRayMath, not floor)
and the gross-vs-net mint-cap behavior above — both discovered via live
execution during verification, not from reading source alone.

Registers ExchangeMetronomeSwap and the MetronomeSwap pool type.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01PtJA4x6iTuhAstcbjb2BoZ


NgoKimPhu:
merged in 06c4a893

PR #1591, opened 2026-08-06
fix: uniswap v3 overquotes thin pools by skipping tick-bitmap words

minhnhathoang:
Quotes for pools holding very little liquidity came out far too high — **8.6x** on the pool that surfaced this — and routes built on them revert on-chain with `Return amount is not enough`. The swap loop jumped straight to the next initialized tick, where the contract walks the empty tick-bitmap words in between one at a time, charging for each.

Fixed by reproducing that step boundary, plus pricing those words in gas so the router stops picking such pools in the first place.

> Rebased onto #1592, which merged the V3 forks into one package. See [Rebase notes](#rebase-notes) at the bottom.

## The pool that surfaced it

Avalanche, Uniswap V4 BTC.b/WETH.e 0.01%.

| | |
|---|---|
| Pool ID | `0x19644199c9a151af99960a39a43686bac3522857b16a8f537a162e791dab0871` |
| PoolManager | [`0x06380C0e0912312B5150364B9DC4542BA0DbBc85`](https://snowscan.xyz/address/0x06380C0e0912312B5150364B9DC4542BA0DbBc85) |
| StateView | [`0xc3c9e198C735a4b97e3e683f391cCBDD60B69286`](https://snowscan.xyz/address/0xc3c9e198C735a4b97e3e683f391cCBDD60B69286) |
| token0 / token1 | BTC.b `0x152b9d0f…3e50` / WETH.e `0x49d5c2bd…0bAB` |
| State | one full-range position, `liquidity = 4787383`, `tickSpacing = 1`, `tick = 291969` |

That works out to reserves of **2 units of BTC.b and 1.05e13 of WETH.e** — roughly five cents of inventory. Pool state: [router-api](https://router-api.kyberengineering.io/avalanche/api/v1/pools?ids=0x19644199c9a151af99960a39a43686bac3522857b16a8f537a162e791dab0871).

**Failing transaction:** https://api.tenderly.co/api/v1/simulations/9a8a0d70-28e2-4222-b54f-0d04392f167a

### Quote comparison — 19 units of BTC.b in

| | amountOut (WETH.e) | |
|---|---|---|
| dex-lib before this PR | 9,331,551,613,090 | 89% of the pool's entire token1 |
| `minReturnAmount` (90%) | 8,398,396,451,781 | |
| **On-chain** | **1,079,518,536,551** | |
| **dex-lib after this PR** | **1,079,518,536,551** | exact |

The chain crossed 9 bitmap words at 1 unit of `amountIn` plus 1 unit of fee each, exhausted the 19 units, and stopped 2178 ticks in.

## The bug

The loop called `nextInitializedTickPos` directly. On-chain, `TickBitmap.nextInitializedTickWithinOneWord` only searches the bitmap word holding the current tick, so a run of empty words is walked one word per iteration. Every stop rounds `amountIn` **up** to a whole unit and charges a fee on top of it, so collapsing that run into a single step lets the input buy far more price movement than it can actually pay for.

That is invisible in ordinary pools, where a word costs vastly more than one unit of input, and total in thin ones — here a word truly costs ~0.025 units but is charged 1 + 1 fee.

## The fix

The primitive was never wrong. The Uniswap/Pangolin SDK has the identical `nextInitializedTick`, but reaches it only through a [`nextInitializedTickWithinOneWord`](https://github.com/pangolindex/sdk/blob/1bf12d68f0b52856d723fb8fd3065050ba6cc57f/src/elixir/utils/tickList.ts#L104) wrapper that clamps the result to the word edge. dex-lib was missing that wrapper, and the SDK's `pool.ts` carries the reason in a comment right above its swap loop:

> `// because each iteration of the while loop rounds, we can't optimize this code (relative to the smart contract)`

So add the wrapper rather than the clamp, and mirror the contract's step boundary inside it.

**One deliberate divergence.** Where the SDK answers an exhausted tick list with the word edge, `ErrBelowSmallest` / `ErrAtOrAboveLargest` keep propagating here. Reaching them means the tick list does not cover the current tick, which in this library signals incomplete tracker data far more often than a genuinely empty bitmap, and walking on at an assumed liquidity would turn missing data into a confident wrong quote.

**`wordCrossings` is dropped.** The old correction modelled only `amountOut` being floored per word and left the `amountIn` round-up — the dominant term — unaccounted for. With steps now spanning at most one word, `ComputeSwapStep`'s own rounding *is* the on-chain rounding.

## Gas

The old model charged only initialized-tick crossings, so this swap — which crosses none — came out as bare base gas, and the pool looked as cheap as it looked deep. Empty-word iterations are now counted and charged `CrossEmptyWordGas`: a cold SLOAD of the bitmap word (2100 under EIP-2929) plus the tick-price math each step runs regardless.

This is a second, independent brake: a swap that walks thousands of words now prices like the multi-million-gas transaction it would really be, so the router deprioritises it on cost alone.

- **Ordinary swaps are untouched** — the UNI/WETH fixture walks no empty words at realistic sizes and still quotes exactly base gas.
- The constant (2500) is an **estimate, not a calibrated measurement**, but it replaces an estimate of zero. Ekubo's independently derived `GasTickSpacingCrossed` is 2507 for the same operation. Happy to retune it against production gas telemetry.
- Wired into the five packages sharing this swap loop; defaults to zero elsewhere, preserving the old model.

## Verification

**On-chain ground truth** — Tenderly simulation `9a8a0d70-28e2-4222-b54f-0d04392f167a`:

```
Swap  amount0        -19
      amount1         1079518536551
      sqrtPriceX96    155352516287591836139164025014435689
      tick            289791
      liquidity       4787383
PoolManager.swap      92773 gas   (9 words, 0 initialized ticks)
```

The simulator now reproduces all of it exactly.

**Independent reference** — expectations come from a separate transcription of the swap loop, not from this simulator, which would have been circular. It reproduces the adjusted values exactly for both directions:

- exact-in: UNI/WETH, 507 steps
- exact-out: solidly-v3 WETH→AI, `157754838261358`

**Test suite** — golden values move by a few units in four suites, always in the direction the quantization biased them: outputs up, exact-out inputs down. `TestTickRoundingToZero`, the regression the old heuristic was written for, still passes. Across all **45 packages** that transitively depend on this math, the failing set is unchanged from `main` (the only failures are pre-existing pancake cases and network-gated tracker tests).

## Cost

| | |
|---|---|
| Swap-path timings | unchanged within noise |
| Allocation count | unchanged, 16/op |
| Allocation bytes | 832 → 864 (`SwapResult` carries the new counter) |
| Complexity | `O(T·log N)` → `O((T+W)·log N)`, `W` bounded by the tick range |
| Marginal cost | ~350 ns per extra iteration |
| Worst case | dust pool walked to saturation, ~1.6 ms |

`W` cannot grow with input, only saturate: 6932 words at tickSpacing 1, 694 at 10, 116 at 60, 36 at 200.

## Note on `uniswapv3-sdk-uint256`

For anyone tempted to delegate this to that SDK instead — it has the same class of bug one layer down. Its `NextInitializedTickWithinOneWord` computes the upward edge as:

```go
maximum := ((wordPos+1)<<8)*tickSpacing - 1        // SDK
(((wordPos+1)<<8) - 1) * tickSpacing               // contract, TS SDK, and this PR
```

For every `tickSpacing > 1` it overshoots by `tickSpacing-1` and lands on a tick that is not even a multiple of the spacing:

| spacing | tick | contract | this PR | SDK |
|---|---|---|---|---|
| 10 | 0 | 2550 | 2550 | 2559 |
| 60 | 0 | 15300 | 15300 | 15359 |
| 60 | −8000 | −60 | −60 | −1 |
| 200 | 0 | 51000 | 51000 | 51199 |

They agree only at `tickSpacing = 1`. This affects `algebra/v1`, `algebra/integral` and `fluid/dex-v2`, which use it — **not addressed here**, separate repo and separate change.

## Rebase notes

Rebased onto #1592 (merging uniswap-v3, pancake-v3, ramses-v2, solidly-v3, slipstream, nuri-v2 into one package). `math.go` was untouched by that refactor; everything else auto-merged.

One conflict needed a decision: #1592 deleted the `solidly-v3` package along with **the only pinned exact-out golden in the tree**, leaving `TestCalcAmountIn`'s round-trip as the sole coverage of a path that rounds the opposite way from exact-in. Since this PR moves that path, the case is carried into the merged package as `TestCalcAmountInSolidlyV3`. Easy to drop if you'd rather it stay deleted.

Bonus: the absorbed forks now share `uniswap/v3`'s `defaultGas`, so they pick up the word-gas term for free.


NgoKimPhu:
@copilot run benchmark tests to check for performance regression

PR #1602, opened 2026-08-10
feat: integrate Everlong CVAMM and CollVault (everlong-cvamm, everlon…

jp1k99ever:
Two config-driven Everlong Labs venues: every pricing parameter is read from
the deployed contracts, so a new chain or venue is a config entry. Adds Katana
chain constants (747474, wrapped native vbETH) for CollVault.

everlong-cvamm: CvammALM — a single-LP AMM on a closed-form curve
(y^2 + (x+c)y - k/x = 0); no pool, ticks or rungs. State is the inventory
coordinate xWad(); band from getSupport(), scale kappa, anchor sqrt price.
Simulator ports CvammCurve.swapExactInX96 wei-exact: closed-form volatile-in,
seeded stable-in bisection, floored conversions, dust fully unspent, partial
fills, solvency clamp, pre-trade directional output fee (read on-chain, not
reproducible off-chain). Exact-input only. Tracker refreshes in one
block-pinned multicall. Verified against a 774-case fixture grid generated
from the shipped CvammCurve bytecode and cross-checked with Everlong's
fuzz-verified Python replica; live CvammQuoter parity test env-gated for the
first deployment.

everlong-collvault: KAI<->vbWBTC settlement venue on Katana priced by the
deployed CollateralRebalancer CR curve via the permissionless Swapper.
Pure-integer port of the deployed math; recovery region refused rather than
mispriced. Wei-exact vs the on-chain quote oracle at pinned block 39284109.

Registers exchange constants, pool types and msgpack bindings.


alex-alra-arteaga:
@ivanpochta  Since the CollateralRebalancer holds the underlying AMM, a swap on the underlying AMM can change the CollateralRebalancer quote. There should be implement coupled pool dependencies/state transitions.

And there's a PR missing in 'ks-dex-adapter-lib', if not, we only support discovery and quoting, not actual aggregator execution.

Would recommend doing it on top of the CollateralRebalancerSwapper.sol

alex-alra-arteaga:
The first issue:
The tracker reads curve state, reserves, supply and withdrawal fee, but not all state used by the actual swap path.
  - Recomputed post-fill reservation value.
  - Post-fill anchor and price-band checks.
  - ICR/MCR safety.
  - Physical-CR safety.
  - For the current workspace version, the leverage interestRate == 0 gate, but I will eventually upgrade it to allow interest rate.

  See src/core/mlv/v1/CollateralRebalancer.sol:491.

Second issue:
The simulator instead applies a hard-coded 2% haircut to one predicted boundary in pool_simulator.go
(https://github.com/KyberNetwork/kyberswap-dex-lib/blob/72bd40226fe5cdad5390dc6a11a7f7254f57bc1d/pkg/liquidity-source/everlong/collvault/pool_simulator.go#L95-L108).
That does not model the other rejection conditions. Its UpdateBalance also leaves PriceWad unchanged and adjusts reference reserves only “approximately.”

Third issue, since interest rates can change the swap without any event, its effect on the debt has to be tracked, these are the venues where it should be tracked:
  - interestRate;
  - the active interest index;
  - the position’s interest index;
  - the last index-update timestamp;
  - projected debt at the quote/execution timestamp.

alex-alra-arteaga:
@jp1k99ever  Not yet now live, but today I will upgrade the CollateralRebalancer to have `hJoin` to be settable, which changes most of the liquidity curve.
Right now the CollVault stores hard-coded Berachain parameters in StaticExtra.
But the exposed params in `leverageCurve()`  in src/core/mlv/v1/CollateralRebalancer.sol:108 will be able to change, it will be a 2 step model, `function setLeverageCurve(uint256 targetCrWad, uint256 hJoinWad) external` sets pending the curve change, `function adoptPendingCurve() external` implements it, and ` function clearLeverageCurve() external` can make the curve return to the default.

Required fix: refresh leverageCurve() into mutable Extra, with a frozen-curve fallback for the currently deployed implementation. The physical CR floor should likewise be read or tied to a verified implementation hash.

Will share the implementation on the morning.

alex-alra-arteaga:
As well the PR README asks for “Allow edits from maintainers”, it's now `maintainerCanModify=false`.
There shouldn't be any mention of Katana/KAI/vbWBTC, it's not supported by Kyberswap.
And the CVAMM live pipeline passed when manually pointed at 0xF512…, but exact quote parity was skipped because no quoter was supplied.

PR #1610, opened 2026-08-17
chore: propagate pool block numbers to DEX metadata

sunspirit99:


NgoKimPhu:
## Code Review

### 🔴 Critical — `wildcard/pool_simulator.go`: `GetMetaInfo` drops native-token flags

**File:** `pkg/liquidity-source/wildcard/pool_simulator.go`

The new implementation returns a bare `pool.MetaInfo` instead of the previous `wildcard.PoolExtra`:

```go
// NEW — native-token flags silently dropped
return pool.MetaInfo{BlockNumber: s.Info.BlockNumber}
```

The old implementation included both `TokenInIsNative` and `TokenOutIsNative`:

```go
// OLD
return PoolExtra{
    TokenInIsNative:  s.IsNative[indexIn],
    TokenOutIsNative: s.IsNative[indexOut],
}
```

`wildcard.PoolExtra` still exists in `types.go` with both fields, and `IsNative` is still populated on the simulator from `Extra`. Any downstream consumer that type-asserts `.(wildcard.PoolExtra)` on the meta info will get the zero value, silently treating all native flags as `false`. Swaps that should send/receive native ETH will be built with WETH instead and will revert on-chain.

**Fix:** Embed `pool.MetaInfo` inside `PoolExtra` and return that, e.g.:
```go
return PoolExtra{
    MetaInfo:         pool.MetaInfo{BlockNumber: s.Info.BlockNumber},
    TokenInIsNative:  s.IsNative[indexIn],
    TokenOutIsNative: s.IsNative[indexOut],
}
```
(Requires adding `MetaInfo pool.MetaInfo` to the `PoolExtra` struct, or re-embedding it.)

---

### 🟠 High — `biswap/pool_tracker.go`: Block number is not pinned to the pool-state snapshot

**File:** `pkg/source/biswap/pool_tracker.go`

The pool state is fetched via `TryAggregate()`, which does not return the block number. The PR compensates with a **separate** `GetBlockNumber` call:

```go
if blockNumber, blockErr := d.ethrpcClient.GetBlockNumber(ctx); blockErr == nil {
    p.BlockNumber = blockNumber
}
```

`GetBlockNumber` returns the **current** chain head at the moment of the call, not the block at which `TryAggregate` executed. On a 2-second block time chain, a new block is very likely to land between the two calls, meaning the stored `BlockNumber` will be ahead of the actual snapshot. Freshness checks and any follow-up reads pinned to this block number will be inconsistent with the pool state.

**Fix:** Switch to `TryBlockAndAggregate` (which returns the block number from the same multicall response) to match the pattern used in all other trackers updated in this PR.

NgoKimPhu (approved):

PR #150, opened 2023-09-19
fix (dodo): invalid error propagation

bbenzikry:

PR #24, opened 2023-06-14
[AG-341] add Dodo simulation logic

it4rb:

PR #21, opened 2023-06-12
[AG-365] add pool simulator for Lido eth-steth

it4rb:
(depends on https://github.com/KyberNetwork/kyberswap-dex-lib/pull/20)

PR #1633, opened 2026-08-27
feat: integrate StonkBrokers Smart Launch V2 launchpad (stonkbrokers-fun-v2)

minhnhathoang:
## Background

**StonkBrokers Smart Launch V2** is a launchpad on **Robinhood Chain** (chainId `4663`). Creators mint a token, the pad seeds a bonding curve, and the curve trades until a graduation market cap closes it and the raise bonds into a locked CL pool.

There is no factory to discover and no creation event to follow: 8 pads are statically deployed, one per quote lane, each hosting many independent launches keyed by a per-pad integer `launchId`.

| Lane | Pad |
|---|---|
| WETH | [`0xFCd61B25BbF3AbD6cf0070D6328E351cc30EEC9f`](https://robinhoodchain.blockscout.com/address/0xFCd61B25BbF3AbD6cf0070D6328E351cc30EEC9f) |
| STONK | [`0x8f6782c5aa37804d08a9b7bf3984ff3245fd6cd4`](https://robinhoodchain.blockscout.com/address/0x8f6782c5aa37804d08a9b7bf3984ff3245fd6cd4) |
| USDG | [`0xd4f20033586977a2511f4a2db4af7c79a340d70a`](https://robinhoodchain.blockscout.com/address/0xd4f20033586977a2511f4a2db4af7c79a340d70a) |
| GME | [`0x4b9dcd6ccfaef0f6d23065dd78e79d5e20ec8cfd`](https://robinhoodchain.blockscout.com/address/0x4b9dcd6ccfaef0f6d23065dd78e79d5e20ec8cfd) |
| NVDA | [`0xee96d955d5634813374ece4c74f2c0ff71b1f9fb`](https://robinhoodchain.blockscout.com/address/0xee96d955d5634813374ece4c74f2c0ff71b1f9fb) |
| AAPL | [`0xb0453a81cbf963903409fff18ad92941e1c7a864`](https://robinhoodchain.blockscout.com/address/0xb0453a81cbf963903409fff18ad92941e1c7a864) |
| SPCX | [`0x0c3b4eded41696eff0ed70841f132b519d81c947`](https://robinhoodchain.blockscout.com/address/0x0c3b4eded41696eff0ed70841f132b519d81c947) |
| USO | [`0xdb3c81c841ff88db6cdfbddb0ee049d162a6053b`](https://robinhoodchain.blockscout.com/address/0xdb3c81c841ff88db6cdfbddb0ee049d162a6053b) |

Shared read/quote lens `SafeLaunchLensV2`: [`0x25b5Df581f4b2Ed450203f375ad8A28b17F115B3`](https://robinhoodchain.blockscout.com/address/0x25b5Df581f4b2Ed450203f375ad8A28b17F115B3). All pads and the lens are verified on Blockscout (solc 0.8.24, cancun, optimizer 200 runs).

`entity.Pool.Address` is a synthetic **`pad#launchId`** composite key — launch ids are per-pad, so the key is not a callable contract. The real pad address travels in `StaticExtra` / `PoolMeta`, the same shape Carbon uses.

## Pricing

Constant product over **virtual** reserves, isolated per launch:

```
tax       = quoteIn * currentTaxBps / 10_000
net       = quoteIn - tax
tokensOut = net * vToken / (vQuote + net)
vQuote   += net ;  vToken -= tokensOut
```

`currentTaxBps` is a flat buffer tax while `now < startTime + bufferSecs`, then decays `decayPerMinuteBps` per minute from `startTaxBps`, flooring at `postTaxBps` (open-ended launches) or 0.

After every buy the pad calls `mcapUsd8(id)` **unconditionally** to decide the graduation close, and that read is strict — so a stale quote-asset oracle reverts the entire buy, not just the graduation decision. The simulator reproduces that. Lanes quote either off a Chainlink-style direct feed or a Uniswap-v3-style TWAP; both paths are ported.

## Scope: buy only

`sell()` caps `tokensIn` to `boughtOf[id][account]`, and `setOperator` hardcodes the grant to `msg.sender` with no permit/signature variant:

```solidity
function setOperator(uint256 id, address operator, bool approved) external {
    if (operator == address(0) || operator == msg.sender) revert BadParam();
    operatorOf[msg.sender][id][operator] = approved;
```

A stateless aggregator route holds no `boughtOf` credit and cannot grant itself an operator approval in the same transaction, so sell is unreachable. `CalcAmountOut` returns `ErrSellNotSupported` for that direction. The vendor docs agree: "Transferred bags cannot be sold through the pad."

## Guards — no quote that reverts on-chain

Every `revert` reachable from `buy() → _launch → _tradeGates → _buy → _splitTax` was enumerated, and each was confirmed by an actual reverting call on a fork.

| Revert | Guard |
|---|---|
| `NotArmed` / `AbortedLaunch` / `AlreadyGraduated` | `ErrPoolNotArmed` / `ErrPoolAborted` / `ErrPoolGraduated` + `ErrPoolBonded` |
| `WindowClosed` | `ErrWindowClosed` |
| `ZeroTrade` (quoteIn 0, net 0, tokensOut 0) | `ErrZeroTrade` |
| `SlippageExceeded` | `ErrSlippageExceeded` |
| `StalePrice` | `ErrStalePrice` / `ErrBadOracleAnswer` |
| **`NotEoa`** | **`ErrEoaOnly`** |
| **`BuyCapExceeded`** | **`ErrBuyCapExceeded`** |

Two deserve attention:

**`eoaOnly` — 74 of 283 live launches (26%).** `_tradeGates` reverts `NotEoa()` when `eoaOnly && msg.sender != tx.origin`. An aggregator swap always calls `buy()` from an executor contract, so those launches are *permanently* unroutable. The flag was read during discovery but never persisted, so they quoted fine and would have reverted on execution.

**`maxBuyPpm` — 11 of 283.** `_buy` credits `boughtOf[id][recipient] += tokensOut` and *then* reverts past `loadedSupply * maxBuyPpm / 1e6`. Neither the simulator nor the pad's own `quoteBuy` applied it, so both quoted trades that cannot execute. Enforced here as a **single-trade upper bound** — see Known limitations for what that bound does not cover.

Not guarded, deliberately: `FeeOnTransferToken` (the quote asset is fixed per pad and all 8 lanes are plain ERC20s; not detectable from state) and transfer reverts inside `_splitTax` (third-party contract behaviour).

## Two chain-level quirks worth knowing

**Robinhood Chain has two independent block-number domains.** The `block.number` opcode and `eth_blockNumber` are separate counters:

```
eth_blockNumber                                    → 46364724
Multicall3.getBlockNumber() (block.number opcode)  → 25837525
eth_call pinned to the opcode-domain number        → "metadata is not found"
```

Multicall-reported block numbers therefore cannot be used to pin dependent `eth_call`s, which silently broke every oracle read. The tracker takes `eth_blockNumber` and pins to that.

Related, and relevant to anyone else integrating on this chain: the public RPC prunes state after roughly 6–8k blocks while the chain produces a block every **~0.101s**, so a pinned refresh has a **~10 minute** shelf life. A slow or rate-limited refresh pass silently degrades pools to untradeable. An archive endpoint removes the problem.

**go-ethereum single-tuple unpacking.** `getLaunch` and `modesOf` each return exactly one `tuple` output, so `Arguments.isTuple()` (`len(args) > 1`) is false and unpacking takes the `copyAtomic` path, which assigns into **field 0** of a struct destination rather than the struct itself. Unpacking straight into `launchRaw` wrote the whole tuple into `launchRaw.Token` (`[20]byte`) and panicked in `abi.setArray`. Fixed with one level of wrapper struct.

## Verification

**Quote correctness** — 32 of 32 quotes matched the on-chain lens **bit-for-bit**, across 7 pools, 3 quote assets, and amounts spanning 1e-06 → 100:

| Case | tax bps | vs lens |
|---|---|---|
| flat 100bps (#176) | 100 | EXACT ×5 |
| flat 200bps (#150) | 200 | EXACT ×5 |
| decay from 9900 (#163) | 100 | EXACT ×5 |
| decay from 1000 (#143) | 100 | EXACT ×5 |
| maxBuyPpm 20000 (#174) | 100 | EXACT ×5 |
| maxBuyPpm 42069 (#120) | 100 | EXACT ×5 |
| GME lane, deep (#24) | 100 | EXACT ×7 |

bps was derived two independent ways — read from the lens, and inverted out of the routed amount via `net = out*vQuote/(vToken-out)` — and the two agreed on every row.

**State transition** — a real `buyEth` executed on a fork at block 46382449 agreed with the simulator on all three quantities:

| | Simulator | Chain |
|---|---|---|
| tokensOut | `23534122942428164868379888` | ERC20 Transfer, identical |
| post-trade vQuote | `420249618506110198` | `getLaunch(176)`, identical |
| post-trade vToken | `975476603161746774578724101` | `getLaunch(176)`, identical |

A full aggregator-routed swap on a fork delivered `23534122942428164868379887`, matching its build quote exactly. `defaultGas` is set from that transaction's measured 405,466 gas rather than estimated.

**Boundaries** — every guard confirmed by a real reverting call:

| Case | On-chain |
|---|---|
| maxBuyPpm 2% (#174), 0.01 | `BuyCapExceeded` |
| maxBuyPpm 4.2069% (#120), 0.1 | `BuyCapExceeded` |
| not armed / aborted | `NotArmed` |
| bonded | `AlreadyGraduated` |
| unknown launch id | `UnknownLaunch` |
| zero amount | `ZeroTrade` |
| eoaOnly (#21), **EOA** sender | succeeds |
| eoaOnly (#21), **contract** sender | `NotEoa` |

**Amounts** on an uncapped launch, 1 wei → 100 ETH: monotonic and correctly asymptotic (100 ETH returns only ~4% more than 10 ETH as the curve approaches `vToken`), and a 1-wei buy still produces non-zero output, so the `ZeroTrade` floor is not hit prematurely.

Note that `eth_call` **cannot** test `eoaOnly`: it sets `tx.origin` to the `from` address, so `msg.sender == tx.origin` holds even when `from` is a contract, and an eoaOnly launch wrongly succeeds. The gate only appears in a genuine nested contract-to-contract call.

Suite: **36 tests passing**, including live integration suites. Live tests dial `STONKBROKERS_RPC_URL` when set and fall back to the public RPC, since the public endpoint rate-limits aggressively.

## Corrections to the vendor documentation

1. `launchIdOfToken` is a **single-argument** call on each **pad** (selector `0xa27d865a`), not a two-argument call on the lens — the lens ABI has 7 functions and does not include it.
2. The floor API's top-level `id` (e.g. `18000176`) is a frontend-synthetic composite; the real on-chain `launchId` is nested at `row.lane.launchId`.

## Smart Launch V3 deliberately excluded

The vendor docs also list 8 **V3** pads for external-token (BYO) launches. They share the same ABI and the same lens, and `externalToken` is written at `createLaunch`, emitted in `LaunchCreated`, and read nowhere else in the pad or in any of its four linked libraries — so it never branches the buy path and adding V3 would be a config-only change.

It is not worth making: **exactly one launch exists across all 8 V3 pads**, and it is `eoaOnly`, i.e. unroutable. `TestV3Pad_ScopeDecision` records the decision and re-checks it against the chain, so a routable V3 launch appearing later will surface.

## Known limitations

- **The `maxBuyPpm` bound is per-trade, not cumulative.** The on-chain check sums `boughtOf[id][recipient]` across a recipient's whole history, which a stateless quote cannot see, so a trade under the bound can still revert if that recipient already holds a position on the launch.
- **`eoaOnly` does not backfill.** Discovery dedupes against already-indexed pools, so the flag only lands for newly discovered launches; an existing index needs clearing for it to apply retroactively.
- **The decay ramp is unexercised live.** Every launch sampled had already passed its window, so tax had floored at `postTaxBps`; the ramp is covered by unit tests only.
- **TWAP lanes.** 59 pools quote off a TWAP. `observe` works at recent blocks; the failures observed during testing were the pinned-block pruning issue described above, not the TWAP port.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,6 +52,22 @@ pkg/liquidity-source/<dex>/
 - Check native token support.
 - Decide whether to reuse an existing integration or create a new one.
 
+### Naming Convention
+
+Name a new source `<protocol>-<family>`. Use that same string for the directory under
+`pkg/liquidity-source/`, for `DexType`, and for the `pkg/valueobject` exchange constant.
+
+- For protocols with multiple generations, prefer nested packages (brownfi/v2, brownfi/v3).
+- Sources outside these families keep a plain name: `carbon`, `pendle`. Existing names stay.
+
+Suffixes:
+- -v2: Uniswap V2 / Solidly / Velodrome forks — e.g. aeon-v2, velodrome-v2
+- -v3: Uniswap V3 forks — e.g. katana-v3
+- -v4: Uniswap V4 / Algebra Integral / hooks — e.g. uniswap-v4-b20
+- -lb: Liquidity Book
+- -prop: Proprietary AMM — e.g. 1010-prop, wasabi-prop
+- -fun: Launchpad bonding curve — e.g. alt-fun, virtual-fun
+
 ### Pool Discovery - IPoolsListUpdater
 
 - Prefer discovering pools on-chain over off-chain indexes.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,6 +52,23 @@ pkg/liquidity-source/<dex>/
 - Check native token support.
 - Decide whether to reuse an existing integration or create a new one.
 
+### Naming Convention
+
+Name a new source `<protocol>-<family>`. Use that same string for the directory under
+`pkg/liquidity-source/`, for `DexType`, and for the `pkg/valueobject` exchange constant.
+
+- For protocols with multiple generations, prefer nested packages (brownfi/v2, brownfi/v3).
+- Sources outside these families keep a plain name: `carbon`, `pendle`. Existing names stay.
+
+
+Suffixes:
+- -v2: Uniswap V2 / Solidly / Velodrome forks — e.g. aeon-v2, velodrome-v2
+- -v3: Uniswap V3 forks — e.g. katana-v3
+- -v4: Uniswap V4 / Algebra Integral / hooks — e.g. uniswap-v4-b20
+- -lb: Liquidity Book
+- -prop: Proprietary AMM — e.g. 1010-prop, wasabi-prop
+- -fun: Launchpad bonding curve — e.g. alt-fun, virtual-fun
+
 ### Pool Discovery - IPoolsListUpdater
 
 - Prefer discovering pools on-chain over off-chain indexes.
--- a/pkg/liquidity-source/stonkbrokers-fun/v2/abi.go
+++ b/pkg/liquidity-source/stonkbrokers-fun/v2/abi.go
@@ -9,9 +9,9 @@ import (
 // PadABI / lensABI are parsed directly from Blockscout's verified-source ABI
 // for the reference pad (0xFCd61B25BbF3AbD6cf0070D6328E351cc30EEC9f) and the
 // shared lens (0x25b5Df581f4b2Ed450203f375ad8A28b17F115B3) -- ground truth
-// pulled during dex-explorer, not the PDF's advertised
+// pulled from the explorer, not the PDF's advertised
 // stonkbrokers.io/integration/ download. All 8 V2 pads share this ABI (same
-// verified contract name + compiler settings; see output/explorer.md).
+// verified contract name + compiler settings).
 //
 // aggregatorV3ABI / twapPoolABI back the buy-side oracle reads (see
 // embed.go).
--- a/pkg/liquidity-source/stonkbrokers-fun/v2/abi.go
+++ b/pkg/liquidity-source/stonkbrokers-fun/v2/abi.go
@@ -1,43 +1,12 @@
 package stonkbrokersfunv2
 
 import (
-	"bytes"
-
-	"github.com/ethereum/go-ethereum/accounts/abi"
+	abiutil "github.com/KyberNetwork/kyberswap-dex-lib/pkg/util/abi"
 )
 
-// PadABI / lensABI are parsed directly from Blockscout's verified-source ABI
-// for the reference pad (0xFCd61B25BbF3AbD6cf0070D6328E351cc30EEC9f) and the
-// shared lens (0x25b5Df581f4b2Ed450203f375ad8A28b17F115B3) -- ground truth
-// pulled from the explorer, not the PDF's advertised
-// stonkbrokers.io/integration/ download. All 8 V2 pads share this ABI (same
-// verified contract name + compiler settings).
-//
-// aggregatorV3ABI / twapPoolABI back the buy-side oracle reads (see
-// embed.go).
 var (
-	PadABI          abi.ABI
-	lensABI         abi.ABI
-	aggregatorV3ABI abi.ABI
-	twapPoolABI     abi.ABI
+	PadABI          = abiutil.MustParseABI(padBytes)
+	lensABI         = abiutil.MustParseABI(lensBytes)
+	aggregatorV3ABI = abiutil.MustParseABI(aggregatorV3Bytes)
+	twapPoolABI     = abiutil.MustParseABI(twapPoolBytes)
 )
-
-func init() {
-	builder := []struct {
-		ABI  *abi.ABI
-		data []byte
-	}{
-		{&PadABI, padBytes},
-		{&lensABI, lensBytes},
-		{&aggregatorV3ABI, aggregatorV3Bytes},
-		{&twapPoolABI, twapPoolBytes},
-	}
-
-	for _, b := range builder {
-		var err error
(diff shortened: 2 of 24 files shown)


minhnhathoang:
`ci / lint` on this PR is red for a reason that is not in this diff: `main` currently fails `ci / lint` on `pkg/liquidity-source/flowstate-c1/pool_simulator.go` (staticcheck QF1008, from #1631), and the PR job lints the merge result.

Fix is up separately as #1634, where `ci / lint` passes. Once that merges this should go green — this PR is already rebased onto latest `main`, and its own package is lint-clean (0 stonkbrokers findings in the CI log).

NgoKimPhu (approved):

PR #1635, opened 2026-08-27
feat(ilyris): add Robinhood Chain discrete-bin AMM

combatwombat1915:
Pull-based discrete-bin AMM on Robinhood Chain. No router: the pool is the swap entry point and the aggregator adapter holds tokens then lets the pool pull (same pattern as `WasabiPropAmmAdapter`, not UniswapV2 transfer-then-swap). Fee is charged once on the input (`netIn = amountIn * (FEE_PRECISION - rate) / FEE_PRECISION`) before any bin is touched; charging per-bin would compound across a traversal and quote low.

Quotes ignore the market guard; swaps do not. `_guardSwap` runs in `swapExactIn` / `swapExactOut` but not in `quoteExactIn` / `quoteExactOut`, so a quote-derived simulator would route into a reverting swap. This adapter models the guard from the same-block snapshot.

> **Addresses corrected 2026-09-02.** The deployment was replaced after this PR was opened and the original description pointed at the superseded generation. The old contracts are still deployed and still answer, so a config built from them indexes a deprecated dust pool without erroring. If you pulled the earlier JSON, please re-copy the block below.

Addresses:

- factory: [`0x3Bf76F2E41Ac7996c822455f4c78fa2026465C4D`](https://robinhoodchain.blockscout.com/address/0x3Bf76F2E41Ac7996c822455f4c78fa2026465C4D)
- lens: [`0x306d65eA97421D33923cCe592dBd22792a789090`](https://robinhoodchain.blockscout.com/address/0x306d65eA97421D33923cCe592dBd22792a789090)
- WETH / wrapped native (already in `wrapped_native.go`): `0x0Bd7D308f8E1639FAb988df18A8011f41EAcAD73`

Superseded, do not use: factory `0x4A943A11a6fFBF8D204Df4d5A080Ca741697ca33`, lens `0xAfd9924C644155dd9384C981c894906EAfeEFd62`.

Two pools are live via this factory: WETH/USDG `0xeEc7cdd6834b4f1feA1D63D32eE0cB7CbB3f24D7` (10 bps bin, 0.01% base fee, guarded) and WETH/YRIS `0xcbb584e1c1259047f2fe03ccb34e8e2086c9f8ad` (250 bps bin, 1.00% fee, **no guard** � `marketGuard()` is the zero address). `allPoolsLength()` is 4; the other two hold nothing and are unreachable through our UI.

Note on the unguarded pool: an `eth_call` to a zero-address guard **succeeds with empty return data** rather than reverting, so a guard read must decode a bare `0x` as "cannot be paused" rather than as a failed read. That distinction broke our own front end before we caught it.

**Source-verified on Blockscout** � factory, lens, both pools, the planner and the market guard. (An earlier revision of this description said verification was incomplete; that was true when written and was fixed the same day.)

No on-chain adapter yet. Pinned-block / archive-fork tests are not included: the public 4663 RPC retains state for only ~6,200 blocks (~10 minutes at ~0.10 s/block), and we have no archive endpoint to offer � our own edge cache buckets historical block tags and is not safe for that purpose. Nothing in this PR needs one; the tests drive a fake chain and make no network calls.

TVL is ~$156. This PR is the code drop, not a listing request.

Config:

```json
{
  "dexID": "ilyris",
  "factoryAddress": "0x3Bf76F2E41Ac7996c822455f4c78fa2026465C4D",
  "lensAddress": "0x306d65eA97421D33923cCe592dBd22792a789090",
  "newPoolLimit": 100
}
```

cc CODEOWNERS @NgoKimPhu @lehainam-dev @SunSpirit


combatwombat1915:
Heads-up: I have edited the PR description. The addresses in the original were superseded by a redeploy after this was opened, and the config JSON in the body pointed at the old factory and lens.

That matters more than a stale address usually would, because the superseded contracts are still deployed and still answer - `allPoolsLength()` on the old factory returns 1, so the old config indexes a single deprecated pool holding a few dollars and never errors. If anyone copied that JSON, please re-copy it.

Corrected in the description: current factory and lens, the two live pools (one of which has no market guard, with a note on decoding that), the state-retention figure for the public RPC, and TVL. I have also stated plainly that source verification on Blockscout is still incomplete for the factory and pools rather than leave the earlier claim standing.

No code changes - the diff is untouched.

combatwombat1915:
Gentle bump. This has been open since 27 Aug and **CI has never run** — I think it is waiting on a maintainer to approve workflows for a first-time contributor, which is why it shows as `blocked` rather than failing. If someone has a moment to click that, it would unblock everything else.

Worth saying so it does not look like a risk: **nothing here needs anything from your side to run.** The adapter's tests drive a fake chain and make no network calls, so there is no RPC, archive endpoint, or fork dependency in CI.

The description is current as of today (see the comment above), and all six of our mainnet contracts are now source-verified on Blockscout, so the code is readable end to end.

Happy to rebase, split it up, or answer anything inline. If you would rather I close and reopen against a different base, just say.

NgoKimPhu:
@combatwombat1915 please follow the repo's convention outlined in AGENTS.md and implement math calculations using uint256.Int instead of big.Int

combatwombat1915:
Thanks @NgoKimPhu — done in the two commits above.

  Math and state are on uint256.Int now. math.go and pool.go have no
  big.Int in code at all; what remains is Info.Reserves and the
  CalcAmountOut signature, which are dex-lib's own types, so the
  conversion happens at that boundary and nowhere else. Second commit is
  goimports -local grouping only.

  One deliberate choice worth flagging: I didn't use big256.MulDiv /
  MulDivUp directly. They call the right primitives but drop the overflow
  flag and the error respectively, and BinPool reverts when a quotient
  won't fit — a silently truncated quote is a route the pool would refuse.
  So MulDiv reads the carry from MulDivOverflow, MulDivUp checks the error
  from MulDivRoundingUpV2, and the running totals use AddOverflow. Happy
  to switch to the helpers if you'd rather be consistent, but it would
  change the failure mode.

  The port is unchanged against the contract: the three oracle tests that
  compare it to a real BinPool on an in-process EVM still pass wei-exact.
  Under CI's own flags the package is clean:

    go test -race -parallel 128 -p 32 -cover -vet=off ./pkg/liquidity-source/ilyris/...
    ok  .../ilyris  1.4s  coverage: 74.6% of statements

  On the red `ci / test` — I don't think it's this PR. In that run ilyris
  reported `ok` at 73.1%; the failure is a data race in gsm-4626, which
  this branch doesn't touch. I reproduced it locally and it's
  deterministic:

    pool_simulator.go:125  CloneState copies CurrentExposure and GhoUsed,
                           but not Rate (*uint256.Int, type.go:19), so a
                           clone shares the original's pointer
    pool_simulator_test.go:101  TestGhoUsedRoundTrip is t.Parallel() and
                                writes cloned.Rate
    pool_simulator_test.go:66   TestCalcAmountOut's subtests write
                                poolSim.Rate

  Same address, two goroutines. It may be failing on main too without
  showing: test_base pipes `go test` into `tee | grep | awk`, so the
  pipeline reports awk's status and a failure gets swallowed.

  Happy to open a separate PR for the CloneState fix if that's useful.

combatwombat1915:
@NgoKimPhu — a correction to my last comment, and one to the description.

**The `ci / test` failure I wrote up had already been fixed before I posted it.** The run I was reading was on `5996e0b4`, where the only failing check was `ci / test`, and the failure was the shared-`Rate` data race in the `gsm-4626` tests. That race was fixed on `main` earlier the same day, in #1647, which merged a few hours before my comment went up — I hadn't refreshed. Sorry for the noise; please ignore that part.

One piece of it does still stand, though: #1647 fixed the test, not the clone. Unless it has changed since, `gsm-4626`'s `CloneState` still copies `CurrentExposure` and `GhoUsed` but not `Rate`, so a clone shares the original's pointer. That one-liner is still on offer if you want it; otherwise I'll leave it alone.

**Description correction.** On 19 Sep we cut the WETH/USDG base fee from 30 bps to 1 bp — `setSwapFee(1)`, i.e. `getBaseFeeRate()` = `100000` against the pool's `FEE_PRECISION` of `1e9`. That is the base rate; the pool's dynamic volatility surcharge applies on top of it. The description said `0.30%`, so I have corrected that line. The addresses in it are unchanged.

The config block is unaffected. Neither `Config` nor `StaticExtra` has a fee field — the only fee the adapter holds is `Extra.TotalFeeRate`, which the tracker re-reads from the lens's `getPoolState` on every refresh, and it never decodes the `swapFeeBps` carried in the `PoolCreated` log. So a fee change on our side needs no config change on yours; it lands on the next tracker tick.

On your uint256 note: the pricing kernel is fully converted — `math.go` and `pool.go` have no `big.Int` in code at all, and what remains is the boundary with dex-lib's own types and the ABI decode structs. The one question I raised there is still open: I hand-rolled `MulDiv`/`MulDivUp` rather than calling `big256`'s, because those drop the overflow flag and the error respectively, and `BinPool` reverts when a quotient will not fit. AGENTS.md asks for the helpers, so if you would rather I be consistent, say the word and I'll switch.

As of writing, the runs on both 8 Sep commits came back `action_required`, so nothing has executed since `5996e0b4`. Whenever you have a moment to approve one, that would show where it actually stands.

PR #1637, opened 2026-08-27
feat(limitorder): cache operator signatures and fetch only the misses

datluongductuan:
## Problem

`GetOpSignatures` calls the limit-order backend for **every** order on **every** RFQ. That endpoint is not read-only — it also writes (`UpdateOpSignatureExpTimeIfActive`), so each redundant call costs a DB write as well as a round trip.

## Why caching here is exact, not an approximation

The operator signs EIP-712 over `(orderHash, opExpireTime)` **only** — never amount, taker or recipient. And the backend (`limit-order` `order_impl.go`, `GetOperatorSignature`) mints a **new** expiry only when the current one has less than `OperatorSigMinimumValidTimeInSecs` left of its `OperatorSigExpTimeInSecs` window — `90s` / `60s` on mainnet.

So two calls for the same orderID inside that ~30s gap return a **byte-identical** signature. Serving one from memory is *equivalent to* calling live, not an estimate of it.

There is production precedent for exactly this: `1inch-resolver`'s `kslimitorder` handler caches per orderID and gates on `cached.expiredAt > nextBlockTime`.

## Design

Caching lives **inside** `GetOpSignatures`, not in a wrapper. That is the whole point: the win is splitting one request into hits and misses and fetching **only the misses**. A wrapper cannot do that — the function is always asked for the full id set.

```
GetOpSignatures(ctx, chainID, orderIds)
  1. reusable(orderIds) -> (cached, uncached)
  2. fetchOpSignatures(ctx, chainID, uncached)   // only the misses hit the network
  3. store(fetched)
  4. return cached ++ fetched
```

The exported contract is unchanged: still `([]*operatorSignatures, error)`, still `nil, nil` on empty data, same error wrapping.

**Reuse predicate.** An entry is served only while it is inside its TTL *and* `expiredAt > now + validityMargin`. The semantically correct bound is "still valid when the tx executes", but dex-lib does not know the target block time — so `validityMargin` stands in for it and defaults to **24s (~2 Ethereum blocks)**. Configurable.

**Storage.** `map[int64]entry` + `sync.Mutex`. Entries are stored **by value** and a fresh pointer is handed to each caller, so concurrent callers never share a mutable `*operatorSignatures`. Past `maxCachedOpSignatures` (16384) unusable entries are swept before new ones are stored; since no entry outlives its TTL, a sweep always reclaims the whole previous window. No LRU, no background goroutine.

**Negative results are not cached.** A dead order is dead forever, but negative caching adds risk for a small win — skipped in v1.

## Config — default OFF

```go
OpSignatureCacheTTL       durationjson.Duration `json:"opSignatureCacheTTL"`       // 0 = cache disabled (default)
OpSignatureValidityMargin durationjson.Duration `json:"opSignatureValidityMargin"` // default 24s
```

This is a shared library. `OpSignatureCacheTTL` defaults to `0`, so **a version bump alone cannot change router-service behaviour** — the consumer has to opt in. `NewHTTPClient` and `NewHTTPClientWithRestyClient` (both deprecated) get no cache at all. Internally the cache pointer is `nil` when disabled and every method is nil-receiver safe, so the disabled path is byte-for-byte the old code path.

## Tests

New `TestGetOpSignatures_Cache` against `httptest.NewServer`, counting the `orderIds` of every request:

| subtest | asserts |
|---|---|
| a repeated orderId is served from cache | 1 request total, identical results |
| only the uncached ids are fetched | requests are `[{1,2}, {3}]`; merged result has all of 1,2,3 |
| a signature expiring inside the validity margin is refetched | 2 requests |
| an entry past the cache TTL is refetched | 2 requests |
| zero TTL disables the cache | 2 requests |
| concurrent callers share one cache | clean under `-race` |

**Verified the tests catch the bug.** With `newOpSignatureCache` neutered to always return `nil` (cache removed), the two cache-behaviour tests fail as intended:

```
--- FAIL: TestGetOpSignatures_Cache/a_repeated_orderId_is_served_from_cache
    expected: [][]string{{"1"}}
    actual  : [][]string{{"1"}, {"1"}}
--- FAIL: TestGetOpSignatures_Cache/only_the_uncached_ids_are_fetched
    expected: [][]string{{"1","2"}, {"3"}}
    actual  : [][]string{{"1","2"}, {"1","2","3"}}
```

The other four subtests still pass without the cache — by construction, since they assert the no-cache request count.

## Checks

- `go build ./...` — clean
- `go vet ./pkg/source/limitorder/...` — clean
- `gofmt -l` / `goimports -local github.com/KyberNetwork/kyberswap-dex-lib` — clean on all three changed files
- `golangci-lint run ./pkg/source/limitorder/...` — 0 issues
- `CI=true go test ./pkg/source/limitorder/... -count=1 -race` — ok
- `go generate ./...` leaves the tree clean (no new simulator registered)

`TestGetOpSignatures_LiveAPI` fails locally with a Cloudflare geo-restriction page from `limit-order.kyberswap.com` ("Access from your country is currently restricted"). That is the endpoint, not this change; the test carries `test.SkipCI(t)` so CI does not run it.

### DIFF
--- a/pkg/source/limitorder/config.go
+++ b/pkg/source/limitorder/config.go
@@ -22,4 +22,11 @@ type Config struct {
 	// HTTPClient lets callers supply their own Transport. http.DefaultTransport keeps only
 	// 2 idle connections per host, which starves callers that fan RFQ out concurrently.
 	HTTPClient *http.Client `json:"-"`
+
+	// OpSignatureCacheTTL caps how long GetOpSignatures may serve an operator signature
+	// from memory instead of asking the limit-order backend. 0 disables the cache.
+	OpSignatureCacheTTL durationjson.Duration `json:"opSignatureCacheTTL"`
+	// OpSignatureValidityMargin is how far past now a cached signature must still be valid
+	// to be reused; defaults to defaultOpSignatureValidityMargin.
+	OpSignatureValidityMargin durationjson.Duration `json:"opSignatureValidityMargin"`
 }
--- a/pkg/source/limitorder/http_client.go
+++ b/pkg/source/limitorder/http_client.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/go-resty/resty/v2"
@@ -19,6 +20,8 @@ const defaultHTTPTimeout = 30 * time.Second
 
 type httpClient struct {
 	client *resty.Client
+	// opSigCache is nil when caching is disabled; every method on it tolerates that.
+	opSigCache *opSignatureCache
 }
 
 // Deprecated: use NewHTTPClientWithConfig, which also supports a timeout, retries and a
@@ -65,7 +68,8 @@ func NewHTTPClientWithConfig(cfg *Config) *httpClient {
 		SetRetryCount(retryCount)
 
 	return &httpClient{
-		client: client,
+		client:     client,
+		opSigCache: newOpSignatureCache(cfg.OpSignatureCacheTTL.Duration, cfg.OpSignatureValidityMargin.Duration),
 	}
 }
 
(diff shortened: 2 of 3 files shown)


datluongductuan:
-

NgoKimPhu (approved):

PR #1656, opened 2026-09-08
feat: integrate prm-fun/uniswap-v4-prm

minhnhathoang:
Adds the MemeCurve bonding-curve pool type and the graduated-pool hook for prm.market's Premium launchpad on Robinhood.

pkg/liquidity-source/prm-fun — ETH-paired MemeCurve pools in Phase.Trading. A virtual-constant-product bonding curve with a 1% ceiling fee and a 4.2 ETH graduation target. QuoteBuy/QuoteSell are transcribed from the Sourcify-verified MemeCurve.sol and pinned to two real on-chain vectors, both reproduced
bit-for-bit (buy: 0.001 ETH -> 747638974590231692804552; sell: 1000e18 ->
1310003014677). Verified end to end on a Tenderly fork: a direct router buy and the same buy routed through the deployed executor both delivered output equal to the pool's own quoteBuy, with native ETH forwarded intact at every hop.

pkg/liquidity-source/uniswap/v4/hooks/premium — once a meme graduates its liquidity moves to a Uniswap v4 pool behind PremiumLaunchHook, so this rides the shared v4 core as a hook overlay rather than a separate pool type, following the uniswap-v4-fables and uniswap-v4-pons-v2 precedents on this chain. Only the 1% desk-token fee delta is DEX-specific.

Notes for reviewers:

- Over-sending past the graduation cap does not revert. The curve caps its own amountIn and refunds the excess, so CalcAmountOut returns RemainingTokenAmountIn to match. Confirmed on-chain: a 10 ETH send against 4.2356 ETH of room consumed only the capped amount.
- A graduated curve's quoteBuy returns zeros rather than reverting, which is a silent-failure trap. The simulator checks phase before any math runs, the tracker parks the pool, and the lister only admits Trading pools.
- A graduated pool is parked with Timestamp=1 so the inactive-pool tracker can archive it, following stonkbrokers-fun/v2's isTerminal convention. The row cannot simply be updated in place: pool-service keys a trading curve by its contract address and the graduated successor by 32-byte v4 poolId.
- Tracker reads decode uint256 into *big.Int, never *uint256.Int. ethrpc v0.7.4 panics rather than erroring on the latter, for both single-value and struct-tuple decodes. Any integration using that pattern is exposed.

Stock-paired pools and the exponential StockCurve are deliberately not included here; they are the subject of follow-up work.


NgoKimPhu (approved):

prm-market:
Hi team, we’ve prepared an update for this PR covering **pHODO-, ETH-, and PRM-paired MemeCurve markets**.

Please check [this branch](https://github.com/prm-market/kyberswap-dex-lib/tree/codex/premium-existing-pr-all-pairs), commit `5d7b13d2023881b5a4e379fa901541ad48f94ec2`. It extends the current PR head (`5269fd89`), resolves the conflicts with main at `7d86da61`, and preserves the existing Premium v4 hook integration.

The update includes pair-aware discovery/backfill, native ETH and ERC-20 handling, and final-fill refunds. All nine local mainnet-fork settlement checks passed: buy, sell, and final buy for each of pHODO, ETH, and PRM. Focused Go tests, race detection, and vet also passed. [Integration notes and reproducible evidence](https://github.com/prm-market/kyberswap-dex-lib/blob/5d7b13d2023881b5a4e379fa901541ad48f94ec2/pkg/liquidity-source/prm-fun/README.md) are included.

Could you review the branch and incorporate the updates into this PR? Our account does not have push access to the Kyber-owned branch. Kyber encoder/production routing still needs your validation. Thanks!

PR #1658, opened 2026-09-09
fix(lo1inch): always parse MakerTraits and implement CloneState

datluongductuan:
Two defects in the `lo1inch` pool simulator, found while re-enabling lo1inch in the vtv2 reserve-taker bot.

## 1. `MakerTraitsInstance` is skipped for empty-extension orders

```go
for _, order := range extra.TakeToken0Orders {
    if order.Extension == "" || order.Extension == helper1inch.ZX { continue }
    order.MakerTraitsInstance = helper1inch.NewMakerTraits(order.MakerTraits)
```

The `continue` fires before the assignment, so an order with no extension keeps `MakerTraitsInstance == nil`. Both fill-amount paths gate the full-fill-only rule on non-nil (`type.go:170` in `CalcTakingAmount`, `type.go:201` in `CalcMakingAmount`), so a `NO_PARTIAL_FILLS` order with no extension is quoted as **partially fillable**. The encoder then builds `fillOrderArgs` with `amount < TakingAmount` and the protocol reverts `PartialFillNotAllowed`.

Fix: the two near-identical loops become one `decodeOrders()`. `MakerTraitsInstance` is assigned first, unconditionally; the extension early-return now only skips the extension-derived instances (`ExtensionInstance`, `FeeTakerExtension`).

This also removed a real divergence between the two loops: the token0 copy swallowed the `NewFeeTakerFromExtension` error behind a commented-out log line, while the token1 copy logged it at Debug. Both now log at Debug.

**Scope honesty:** in a 24h sample of 3,420 on-chain fills the `(NO_PARTIAL_FILLS, HAS_EXTENSION)` split is (0,0):1701, (1,1):925, (0,1):794 — **zero** orders with `NO_PARTIAL_FILLS=1` and `HAS_EXTENSION=0`, which is the only combination that triggers the bug. So this is not currently causing reverts in that window. It is still a real defect: the protocol does not couple those two bits, and the same fix covers the `HAS_EXTENSION`-with-empty-extension case that would otherwise mis-encode silently.

## 2. `CloneState()` is not implemented

The simulator does not override `CloneState()`, so it falls through to `pool.Pool.CloneState()`, which returns `nil`. `Order` is a pointer shared with the entity-derived slices and with the `orderHash -> index` maps, so a caller doing clone-then-`UpdateBalance` mutates the shared simulator's orders instead of a copy. reserve-taker's backrun path clones pending-pool simulators, so this silently corrupts normal-flow state.

Fix: `CloneState()` plus a `cloneOrders` helper — `cloned := *p`, then both order slices rebuilt with a fresh `*Order` per element.

`minBalanceAllowanceByMakerAndAsset` is deliberately **not** cloned. `UpdateBalance` does not write it: it writes `order.{RemainingMakerAmount,MakerBalance,MakerAllowance}` and the external `params.SwapLimit`. The map is written only in `NewPoolSimulator` and read only in `CalculateLimit`, which builds fresh `*big.Int` via `ToBig()`. Per `AGENTS.md:82` ("deep-copy every slice/map/pointer `UpdateBalance` writes by index or in place") sharing is correct and conventional, and it avoids an O(orders) map allocation on every clone in the backrun hot path, where books run to pool-service's 500-order cap. Same reasoning for the two hash->index maps and `Info.Reserves`.

## Tests

Verified before/after by stashing only `pool_simulator.go`: **0 passed / 3 failed** before, **3 pass** after.

| Test | Before |
|---|---|
| `TestPoolSimulator_CalcAmountOut_NoPartialFillsWithoutExtension` | partial take is quoted, so `require.ErrorIs(ErrCannotFulfillAmountIn)` fails with "got nil" |
| `TestPoolSimulator_NoPartialFillsSurvivesMsgpack` | same, through `EncodePoolSimulatorsMap`/`DecodePoolSimulatorsMap` — confirms `IncludeUnexported(true)` carries `MakerTraitsInstance` and its unexported `*big.Int` through the transport router-service actually uses |
| `TestPoolSimulator_CloneState` | `require.NotNil(sim.CloneState())` fails — base returns nil |

The defect-1 fixture uses real mainnet `makerTraits` `0x8a00000000000000000000001254000014006a9f8ac700000000000000000000` (maker `0x9a0ef3c3…`, `NO_PARTIAL_FILLS=1`, `ALLOW_MULTIPLE_FILLS=0`, bit-invalidator) with only the expiration cleared via `helper.WithExpiration(nil)`, so the test does not depend on the wall clock.

## Deliberately out of scope

**`AllowedSender` is not enforced here**, and I recommend it stays out. `AllowedSender()`, `IsPrivate()`, `IsBitInvalidatorMode()` and `IsEpochManagerEnabled()` have zero call sites outside `helper/maker_traits.go`, so the simulator does quote orders it cannot legally fill — but this is the wrong layer to fix it:

1. `p.takerAddress` comes from `staticExtra.TakerAddress`, written by the pool-service scanner. Where that differs from the address that is actually `msg.sender` at the protocol, enforcement would silently delete private orders from other consumers' quotes. The existing fixtures have no `takerAddress` at all (zero address), which would drop every private order.
2. Wrong semantics for that field: here `takerAddress` is the **fee** identity (`FeeTakerExtension` -> `Whitelist.IsWhitelisted`, a half-address match), while the protocol checks `allowedSender` against `msg.sender`.
3. It is the only *static* member of a family of live-state predicates — bit-invalidator consumption, epoch, expiry-at-execution-block, remaining staleness. Enforcing just this one creates a false impression of legality checking.
4. The caller that needs it is building a chain-state overlay anyway, where `!IsAllowedSender(actualMsgSender)` is one line in the same pre-flight.

If it is ever wanted here, the honest shape is an opt-in `StaticExtra` field defaulting to off.

Also noticed, untouched: `rfq.go` returns `(nil, nil)` from both `RFQ` and `BatchRFQ` — a no-op handler that reports success; `Order.GetFilledMakingAmount()` always returns 0 and `GetRouterContract()` returns `""`, both looking like unimplemented stubs; and `helper/whitelist.go` already fails `gofmt -l` on `main` (verified by stashing).

## Verification

`go build ./...` · `go test -race ./pkg/liquidity-source/lo1inch/... ./pkg/msgpack/...` · `golangci-lint run ./pkg/liquidity-source/lo1inch/...` — all pass.

### DIFF
--- a/pkg/liquidity-source/lo1inch/pool_simulator.go
+++ b/pkg/liquidity-source/lo1inch/pool_simulator.go
@@ -68,52 +68,12 @@ func NewPoolSimulator(entityPool entity.Pool) (*PoolSimulator, error) {
 		return nil, err
 	}
 
-	for _, order := range extra.TakeToken0Orders {
-		if order.Extension == "" || order.Extension == helper1inch.ZX {
-			continue
-		}
-
-		order.MakerTraitsInstance = helper1inch.NewMakerTraits(order.MakerTraits)
-
-		extensionInstance, err := helper1inch.DecodeExtension(order.Extension)
-		if err != nil {
-			return nil, fmt.Errorf("decode extension: %w", err)
-		}
-		order.ExtensionInstance = &extensionInstance
-
-		feeTakerExtension, err := helper1inch.NewFeeTakerFromExtension(extensionInstance)
-		if err != nil {
-			// logger.Errorf("failed to decode fee taker extension: %v", err)
-			// not always that extension data can be used to create new fee taker extension
-			// so we need to continue
-			continue
-		}
-
-		order.FeeTakerExtension = &feeTakerExtension
+	if err := decodeOrders(extra.TakeToken0Orders); err != nil {
+		return nil, err
 	}
 
-	for _, order := range extra.TakeToken1Orders {
-		if order.Extension == "" || order.Extension == helper1inch.ZX {
-			continue
-		}
-
-		order.MakerTraitsInstance = helper1inch.NewMakerTraits(order.MakerTraits)
-
-		extensionInstance, err := helper1inch.DecodeExtension(order.Extension)
-		if err != nil {
-			return nil, fmt.Errorf("decode extension: %w", err)
-		}
-		order.ExtensionInstance = &extensionInstance
-
-		feeTakerExtension, err := helper1inch.NewFeeTakerFromExtension(extensionInstance)
-		if err != nil {
-			logger.Debugf("failed to decode fee taker extension: %v", err)
-			// not always that extension data can be used to create new fee taker extension
-			// so we need to continue
-			continue
-		}
-
-		order.FeeTakerExtension = &feeTakerExtension
+	if err := decodeOrders(extra.TakeToken1Orders); err != nil {
+		return nil, err
 	}
 
 	takeToken0OrdersMapping := make(map[string]int, len(extra.TakeToken0Orders))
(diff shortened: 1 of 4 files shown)


datluongductuan:
CI here is red for a reason unrelated to this PR: `main` itself does not build.

`pkg/source/limitorder` declares `opSigServer` twice — #1628 added it as a function in `rfq_test.go`, #1637 later added a struct of the same name in `http_client_test.go`, and the collision only appears once both are on the same branch. I reproduced it on a clean `origin/main` worktree, and this PR does not touch that package at all (it only touches `pkg/liquidity-source/lo1inch`).

Fix is up as #1661 — a two-line test-only rename. Once that merges this PR should go green on a rerun; happy to rebase if you prefer.

NgoKimPhu (approved):

datluongductuan:
Rebased onto current `main` and folded in the CI fix, so this should now go green on its own.

Two things happened here:

**The branch was stale.** It was based on `39d27f2a`, a `main` old enough that neither `rfq_test.go` nor the `opSigServer` struct in `http_client_test.go` existed on it yet. The collision only ever appeared in CI, which tests the merge with current `main` — which is also why it looked like this PR's fault when it is not. Now rebased onto `77447f74`.

**The collision fix came with it.** `opSigServer` was declared twice in `pkg/source/limitorder`: #1628 added it as a function, #1637 later added a struct of the same name. Neither was wrong on its own branch. The struct keeps the name — it is the richer helper and carries methods — and the function becomes `signedOrderServer`, which also describes it better. Test-only, four lines. #1661 is closed as redundant.

Note this means `main` is still broken for every *other* open PR until this one merges.

Verified after the rebase: `go build ./...` clean, and `go test ./pkg/source/limitorder/ ./pkg/liquidity-source/lo1inch/... ./pkg/msgpack/...` → 181 pass across 9 packages.

One pre-existing thing this PR does **not** touch and does not fix: `pkg/liquidity-source/lo1inch/helper/whitelist.go` fails `gofmt -l` on `main` already — confirmed by formatting the file as it stands on `main`, and it is absent from this PR's diff.

PR #1661, opened 2026-09-11
fix(limitorder): rename the colliding opSigServer test helper

datluongductuan:
`main` does not build, so every open PR inherits a red CI.

`pkg/source/limitorder` declares `opSigServer` twice. #1628 added it as a **function** in `rfq_test.go`; #1637 later added a **struct** of the same name in `http_client_test.go`. Neither PR was wrong on its own branch — the collision only exists once both are on `main`, which is why it slipped through.

```
pkg/source/limitorder/rfq_test.go:19:6: opSigServer redeclared in this block
	pkg/source/limitorder/http_client_test.go:144:6: other declaration of opSigServer
pkg/source/limitorder/rfq_test.go:56:24: too many arguments in conversion to opSigServer
```

Reproduced on a clean `origin/main` worktree, so this is not specific to any branch.

The struct keeps the name: it is the richer helper, it carries methods (`newOpSigServer`, `recorded`), and renaming it would touch four sites instead of two. The function becomes **`signedOrderServer`**, which also describes it better — it serves signatures only for the ids the caller marked as signed.

Test-only, no production code touched. `go vet ./pkg/source/limitorder/` clean, `go test ./pkg/source/limitorder/` 93 pass, `gofmt -l` clean.

### DIFF
--- a/pkg/source/limitorder/rfq_test.go
+++ b/pkg/source/limitorder/rfq_test.go
@@ -14,9 +14,9 @@ import (
 	"github.com/KyberNetwork/kyberswap-dex-lib/pkg/source/pool"
 )
 
-// opSigServer answers /orders/operator-signature with a signature for every requested id
-// present in signed, mimicking the service dropping ids that are no longer active.
-func opSigServer(t *testing.T, signed map[int64]bool) *httptest.Server {
+// signedOrderServer answers /orders/operator-signature with a signature for every requested
+// id present in signed, mimicking the service dropping ids that are no longer active.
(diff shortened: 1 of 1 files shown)


datluongductuan:
Folded into #1658 instead of landing separately, at the author's request — the rename now rides along there, so this PR is redundant.

Same two-line change, same reasoning: the struct in `http_client_test.go` keeps the name (it is the richer helper and carries methods), and the function in `rfq_test.go` becomes `signedOrderServer`.

PR #1668, opened 2026-09-13
fix: uniswap-v4 Track sees this refresh's pool state, not the previous one's

lehainam-dev:
## Why did we need it?

`hook.Track` sees the **previous** refresh's pool state. `fetchOnchainState` passes it the pre-refresh entity while its own `eth_call`s run at the newly fetched block, and the callers only write this refresh's liquidity/slot0/ticks into `p.Extra` after `resolveHookState` returns. #1540 fixed this ordering for `Reserves`, but not for `Extra`.

For auto-calibrated pools this puts the fitted fee on the wrong swap side, which can quote a zero fee and overquote. E.g. on pool `0xdbed57e1dc571d912097e4dd1808698c7d4f4cc9647a4be0b05f8ff57c92c317` (Robinhood), 0.16 WETH quoted `402590010` USDG against `392530678` on-chain.

**Fix:** `resolveHookState` writes this refresh's liquidity, slot0 and ticks into `Pool.Extra` before `hook.Track`. `hX` is left out: hooks read their state from `HookParam.HookExtra`, and the callers persist `Track`'s output afterwards. This covers `BootstrapPoolState` and `updateState` (used by `GetNewPoolState` and `GetNewPoolStateWithOverrides`).

## Related Issue

Follow-up to #1540.

## Release Note

- No config or schema changes.
- On the overrides path, a `Track` error now returns the pool with the refreshed `Extra`, as it already did for `Reserves`.
- `pancake-infinity/cl` has the same ordering; left for a separate change.

## How Has This Been Tested?

- New `TestResolveHookState_TrackSeesFreshPoolState`: fails on `main`, passes with the fix.
- `CI=1 go test ./pkg/liquidity-source/uniswap/v4/...`: 20 packages ok, 0 failed.
- Not yet validated end to end through pool-service.


pkg/liquidity-source/uniswap/v4/pool_tracker.go
```diff
@@ -156,6 +158,26 @@ func (t *PoolTracker) resolveHookState(
...
 
+// refreshPoolExtra writes this round's slot0, liquidity and ticks into p.Extra
+// before Track runs. Callers persist the fresh Extra only after
+// resolveHookState returns, but Track's RPC calls observe the new block, so a
+// hook reading Pool.Extra in Track would otherwise see the previous round's
+// state. The existing hX is kept.
+func refreshPoolExtra(p *entity.Pool, result *FetchRPCResult, ticks []Tick) {
+	var extra Extra
+	_ = json.Unmarshal([]byte(p.Extra), &extra) // best effort: a new pool has no Extra yet
+	extra.Extra = &uniswapv3.Extra{
```
NgoKimPhu: I think it's better to assign field by field (or create a method to copy at struct definition) to avoid missing copying new fields when they are added in the future

NgoKimPhu: I see that this is copied from around line 866. A helper maybe?

pkg/liquidity-source/uniswap/v4/pool_tracker.go
```diff
@@ -156,6 +158,26 @@ func (t *PoolTracker) resolveHookState(
...
 }
 
+// refreshPoolExtra writes this round's slot0, liquidity and ticks into p.Extra
+// before Track runs. Callers persist the fresh Extra only after
+// resolveHookState returns, but Track's RPC calls observe the new block, so a
+// hook reading Pool.Extra in Track would otherwise see the previous round's
+// state. The existing hX is kept.
+func refreshPoolExtra(p *entity.Pool, result *FetchRPCResult, ticks []Tick) {
+	var extra Extra
+	_ = json.Unmarshal([]byte(p.Extra), &extra) // best effort: a new pool has no Extra yet
```
NgoKimPhu: Is there a way we can avoid Unmarshalling twice? For other dexes it's ok but let's optimize for uniswap

pkg/liquidity-source/uniswap/v4/pool_tracker.go
lehainam-dev: Moved to a method in rpcData struct definition.

NgoKimPhu (approved):

PR #1697, opened 2026-09-19
feat(uniswap-v4): support Arcade launchpad hook v2 (Arc)

obseasd:
## Why did we need it?

Follow-up to #1674 (`uniswap-v4-arcade`, the Arcade launchpad hook on Arc) and #1675 (`arcade-fun`, its bonding curve). Arcade has deployed a second generation of the hook on Arc mainnet (chain 5042), and new launches will be made on it. This PR makes both sources serve v2 next to v1. **v1 behaviour is untouched.**

**ArcadeHook v2**: `0x7706d261f0C370e8f0E273164A603E885C89beC2`, deployed at block 21568825, source verified: https://arc.etherscan.io/address/0x7706d261f0C370e8f0E273164A603E885C89beC2#code

- Permission bitmap `0x3EC2`. v1 (`0x695cfF9C7F11fa87ca05c7b0A0fa64C3554B3eCe`) is `0x3ECE`: v2 is v1 without `BEFORE_SWAP_RETURNS_DELTA` and `AFTER_SWAP_RETURNS_DELTA`. The hook no longer does any custom accounting on a swap.
- Same PoolKey shape as v1: sorted currencies (the launch token and `quoteAssetOf(token)`, zero address meaning USDC), `fee = poolFeeOf(token)`, `tickSpacing = 200`, `hooks = the hook`, empty hookData. Same canonical PoolManager (`0x8366a39CC670B4001A1121B8F6A443A643e40951`).
- Status today: the hook is deployed and verified, but its Safe-owned wiring batch is not executed yet, so **there is no pool on it yet**. Pools will appear once launches start, within days. Nothing to index until then; this PR is so the first v2 pool is priced correctly from its first block.

**What changed for a swap**

- **PUMP, graduated.** On v1 the pool's LP fee is 0 and the hook takes the whole fee in USDC as a swap delta (1% decaying to 0.30% from the `feeObs` oracle, the `MAX_TOTAL_TAKE` clamp). On v2 the pool carries a **static native LP fee of 10_000 (1%) in its PoolKey from creation**, `beforeSwap` returns `ZERO_DELTA` and `afterSwap` returns no hook fee. The V4 pool math with the pool's own fee is the whole quote.
- **`feeObs` is gone from v2** (so is the `SwapTreasuryFee` event, which this lib never read). The getter does not exist, the call reverts, and inside `Aggregate()` it would fail the whole batch and leave every v2 pool untracked. `Track` skips it on v2.
- **CLANKER / RWA**: as on v1. Native LP fee, and the per-transaction buy cap during the first five minutes (1% of supply in the first minute, +1% per minute).
- **Curving PUMP**: as on v1, V4 swaps revert while `status != Graduated`; the simulator refuses them.
- **Exact-out**: v1 stays exact-in only, as merged. A hook that never returns a swap delta has no fee side to get wrong, so `CalcAmountIn` is allowed on v2 pools.

**Changes**

- `pkg/liquidity-source/uniswap/v4/hooks/arcade`
  - `HookV1`, `HookV2`, and `Generations`, a map from hook address to the flags that matter for a swap (`NoSwapDelta`). The generation is decided **from the hook address, never from RPC**: the factory sets it from `HookParam.HookAddress` whatever the persisted extra says. The zero value is v1, so a pool whose flag went missing would be quoted with the v1 hook fee: too low, never too high.
  - `BeforeSwap` / `AfterSwap` take the PUMP fee only on a generation that charges it in the hook. `Track` reads `feeObs` only there.
- `pkg/liquidity-source/arcade/curve` (`arcade-fun`)
  - The curve (`hook.buy` / `hook.sell`) is unchanged in math and interface on v2 but lives at the new hook address. The source pinned one hook through `Config.Hook`; the config now also takes `hooks`, merged with `hook`, so an existing single-hook config keeps working.
  - A launch is stamped with the hook that emitted its `LaunchCreated` (`StaticExtra.hook`), and the tracker reads, the approval address, `SwapInfo.Hook` and `MetaInfo.ApprovalAddress` all follow that per-pool hook. PoolIds hash the hook address, so v1 and v2 launches can never share a synthetic pool address.

## Related Issue

Follow-up to #1674 and #1675, same integration.

## Release Note

- `uniswap-v4-arcade`: nothing to configure, the v2 hook is registered by address like v1.
- `arcade-fun`: add the v2 hook to the source config (and to whatever feeds it the hook's logs), for example:
```json
{ "dexId": "arcade-fun", "chainId": 5042, "hooks": ["0x695cff9c7f11fa87ca05c7b0a0fa64c3554b3ece", "0x7706d261f0c370e8f0e273164a603e885c89bec2"] }
```
  The previous `{"hook": "0x695c..."}` form still works and serves v1 only. For execution, the contract to approve and to call `buy` / `sell` on is per pool (`StaticExtra.hook`, surfaced as `MetaInfo.ApprovalAddress` and `SwapInfo.Hook`), now one of two addresses.
- No new exchange, pool type or msgpack registration.

## How Has This Been Tested?

- **v2 parity (`TestDirectProtocolMatch_GraduatedPumpV2`)**: the full `uniswapv4.PoolSimulator` with this hook against outputs of the real v2 hook (address bits `0x3EC2`) executed on a real v4-core PoolManager in Foundry. There is no v2 pool on mainnet yet, so this is the only source of real outputs. A PUMP launch is graduated into its full-range locked position (pool fee 10000, read back from `slot0`), moved off its graduation price by swaps across time, then each swap is executed from a state snapshot: 4 buys and 3 sells exact-in, 3 buys and 3 sells exact-out, for both currency orderings.
  - Inside one tick-bitmap word every case is within the same 25-unit band as the v1 fixtures (measured 0 to 21), always on the safe side: never more output than the contract pays, never less input than it charges.
  - **One finding, inherited and not specific to this hook.** Three of the large buys move the price across a tick-bitmap word boundary (256 x tickSpacing = 51,200 ticks). `PoolManager.swap` ends a step at every word boundary, initialized tick or not, and rounds that step's input and fee up; the shared swap math steps straight to the next initialized tick (`NextInitializedTickIndex`), which on a full-range-only pool is never reached. On those three cases the contract pays less than quoted by the output worth of 0.5 to 1.3 **input base units** (about 1e-6 USDC; 2.5e-10 of the output at most). I replayed all the fixtures with an exact port of `SwapMath` plus the word stepping and it reproduces every contract output to the unit, which is how the gap is attributed. The test asserts it explicitly (`maxWordStepInputUnits`, 2 input units per word crossed) instead of choosing sizes that avoid it. It would apply equally to v1 pools and to any v3 / v4 pool whose liquidity is one wide range; the v1 fixtures simply never crossed a word. Happy to look at it in the shared math separately if you want it closed.
- **v1 unchanged**: every existing test and fixture of #1674 passes unmodified (`TestPumpFeeBps`, the buy / sell fee sides, `TestDirectProtocolMatch_GraduatedPump`, the V4Quoter-verified CLANKER fixture).
- **Unit tests added (`hook_test.go`)**: the address to generation map and the permission bits of both addresses; the factory trusting the address over a persisted extra; zero hook fee on v2 PUMP for both orderings, both directions, exact-in and exact-out; the buy cap on v2 CLANKER / RWA including exact-out; the curve guard on v2; exact-out still refused on v1.
- **`Track` against a stubbed Multicall3 (`hook_track_test.go`, runs in CI)**: a stub that reverts `aggregate` when `feeObs` is called, as the v2 contract does. v2 never asks for it and decodes; v1 still reads the oracle (fee 78 bps from the #1674 fixture values); and the v1 batch pointed at a v2 stub fails, which shows the stub is faithful. Mutation check: registering v2 with the v1 flags makes `TestGenerations`, `TestTrack_FeeObsOnlyOnV1` and the v2 parity test fail.
- **Live, skipped in CI**: `TestTrack_LiveV2` and `TestGetNewPoolState_LiveV2` run both trackers against the deployed v2 hook on Arc mainnet. With no launch yet the mappings read as zero values; what they show is that the batches decode on the real contract (with `feeObs` in it the call reverts, checked with a direct `eth_call`). `getHookPermissions()` returns 16066 = `0x3EC2` on v2 and 16078 = `0x3ECE` on v1.
- **Curve (`arcade-fun`)**: `ArcadeV4Curve.sol` is byte-identical between the two hooks and `sell` is unchanged; `buy` differs only in an argument of the graduation call. To back that with execution, all 9 parity vectors of #1675 (3 fresh buys, 2 mid-curve buys, the clipped graduating buy with its refund, 3 sells) were re-executed against the v2 bytecode in Foundry: bit-exact. `TestPoolFactoryDecoder_BothHooks` covers discovery from both hooks under three config shapes, foreign emitters, and a legacy single-hook config ignoring the other hook; `TestV2Launch_SameCurveOwnHook` covers pricing and the per-pool approval / call target.
- **Build**: `go build ./...` and `go vet` on the touched packages are clean (go 1.25.10). `go test` passes for `./pkg/liquidity-source/uniswap/v4/hooks/arcade/`, `./pkg/liquidity-source/arcade/curve/`, `./pkg/liquidity-source/uniswap/v4/`, `./pkg/valueobject/` and `./pkg/pooltypes/`. `go generate ./pkg/msgpack/` registers the same set of types.
- **CI**: `test`, `test_base` and the generated-code check pass. `ci / lint` is red on `main` as well; the two findings left are in `pkg/liquidity-source/lunya-fun` (`ineffassign`, `unused`) and are not touched here. This PR adds none, and clears the one `SA1012` (nil context) that the #1675 test had.

v1 source is public at https://github.com/obseasd/arcade-integrations, v2 is on the Etherscan link above. Happy to share the Foundry fixture generators or run more quotes on request.

https://claude.ai/code/session_01NLjP6noF6zmsWFKoZXWeT3


obseasd:
Friendly ping on this one, it has been open since 19 Sept with no reviewer assigned.

Some context on why it matters on our side: this is not a new integration but the second generation of the hook you already merged in #1674. Arcade redeployed the hook on 19 Sept as `0x7706d261f0C370e8f0E273164A603E885C89beC2` (the v1 `0x695cfF9C...` keeps the pools it already hosts), and a Uniswap V4 hook cannot be upgraded in place because its permission bits are encoded in its address. So every pool created on Arc from now on lives behind the v2 address, and until this merges KyberSwap routes only the previous generation.

The v2 deliberately drops `beforeSwapReturnsDelta` and `afterSwapReturnsDelta` (bits `0x3EC2` instead of `0x3ECE`) and takes a static pool fee instead, which should make it simpler on your side than v1, not harder.

On CI: `ci / lint` is red here, but it is red on `main` too and on every PR merged since 17 Sept, including our own #1675. The two findings are `pkg/liquidity-source/lunya-fun/abi.go:14` (`tradeEvent` unused) and `curve.go:81` (ineffectual assignment to `snipeBps`), neither of them in files this PR touches. Happy to open a separate one-line PR fixing those two if that unblocks your pipeline, just say the word. `ci / test`, `ci / test_base` and the generated-code check all pass.

Anything you need changed in the adapter, I will turn it around quickly.

PR #1704, opened 2026-09-22
feat(uniswap-v4): export DefaultGas

lehainam-dev:
## Why did we need it?

`uniswapv4.defaultGas` is the gas the V4 simulator charges per swap (`BaseGas`) and per crossed initialized tick (`CrossInitTickGas`). It's unexported, so code outside the package that needs the same numbers has to copy them.

For example, code that estimates a hook's extra gas on top of a plain V4 swap has to subtract what the simulator already charges for crossed ticks. Exporting the value keeps such callers in sync if it changes here.

It can't use the V3 simulator's `Gas` instead: `uniswapv3`'s default is `CrossInitTickGas: 21492`. `NewPoolSimulator` overwrites that with this package's value.

Change: `defaultGas` → `DefaultGas`, the same convention as `DefaultGas` in `woofi-v2`, `curve/*`, `balancer/v2/*`, and `dodo`. Values are unchanged.

**Also fixes the 3 lint errors already on `main`** (CI `ci / lint` failed on them, not on this change):
- `lunya-fun/abi.go`: removed the unused `tradeEvent` (assigned in `init`, never read).
- `lunya-fun/curve.go` (`quoteBuy`): removed the dead `snipeBps = feeBps - c.curveFeeBps` inside the fee cap. Only the capped `feeBps` is read afterwards, so the output is unchanged. @anton-erwang: if the snipe/curve fee split was meant to be reported separately, that's a missing use, not dead code. Please check.
- `arcade/curve/pool_simulator_test.go`: `context.Background()` instead of a `nil` Context.

## Related Issue

N/A

## Release Note

Additive: one newly exported identifier, no behavior change. It was only referenced inside `uniswap/v4`, and both references are updated.

## How Has This Been Tested?

- `golangci-lint run` on `arcade/...`, `lunya-fun/...`, `uniswap/v4`: 0 issues (local v2.11; CI uses v2.4). `arcade/curve` and `lunya-fun` tests pass.
- `go build ./...` and `go vet ./pkg/liquidity-source/uniswap/v4/` pass.
- `go test ./pkg/liquidity-source/uniswap/v4/...` gives the same results with and without this change (stash vs. apply). The only failures are live-RPC tests (`TestPoolTracker_*`, `*_Track`, `*_Live`, bunni-v2 `Test_Quoter`). From my network the RPC returns 403 "access restricted", and which of the live tests fail varies between runs.


pkg/liquidity-source/arcade/curve/pool_simulator_test.go
```diff
@@ -231,7 +232,7 @@ func TestPoolFactoryDecoder(t *testing.T) {
 	assert.Nil(t, p)
 
 	for _, ev := range []common.Hash{curveBuyEventHash, curveSellEventHash, graduatedEventHash} {
-		addrs, err := d.DecodePoolAddressesFromFactoryLog(nil, types.Log{Address: common.HexToAddress(testHook), Topics: []common.Hash{ev, poolID}})
+		addrs, err := d.DecodePoolAddressesFromFactoryLog(context.Background(), types.Log{Address: common.HexToAddress(testHook), Topics: []common.Hash{ev, poolID}})
```
lehainam-dev, line 235: Review note: Fix lint err.

NgoKimPhu (approved):
Export is correct and low-risk (verified NewPoolSimulator overwrite, and that removing lunya-fun's snipeBps reassignment doesn't change quoteBuy's output — it isn't returned). Bundling the 3 unrelated lint fixes into this PR is a bit of scope creep, but they're disclosed and trivial, not blocking.

